PR #8, opened 2022-11-10
Fix sniper assassin contracts for co-op

scrungofan:
The old contracts were missing the multiplayer bricks, and the character images for falcon. This fixes that.


RDIL (approved):
LGTM

PR #10, opened 2022-11-10
Fixed flags from being loaded rather late in the startup phase

LennardF1989:
Because imports can also use flags, and there are a lot of imports before the server even starts, it makes sense to load those as soon as possible to prevent undefined errors.

In this PR, that is fixed as well as adding a toggle for LiveSplit (which is secretly instanced and tried during the startup phase), so it won't try to make the connection at all (and thus error out if it's not available).


fu5ha (approved):
nice, LGTM :)

PR #14, opened 2022-11-12
Add H2 Escalations

Kakiking:


AnthonyFuller (changes requested):
Great work, but there's an issue, a big issue.

None of these have entitlements, I presume these were retrived from Hitman 2's servers, and the way Hitman 2 works is that there aren't location entitlements (from memory), but they need to be added _for Hitman 3_, Peacock will automatically transform them for Hitman 2.

`MIAMI -> SGAIL`: `H2_LEGACY_STANDARD`
`HAVEN -> NEW YORK`: `H2_LEGACY_EXPANSION`

Kakiking:
oh, ok, time to go over every json again and add those... Thanks

AnthonyFuller (approved):
LGTM

PR #23, opened 2022-11-13
Fixed issue where the newEvent-hook is called twice

LennardF1989:
Currently, the newEvent-hook is called twice. This PR fixes that issue by removing the second instance. Reason being the first instance is very early in the process of handling an event, giving plugins a chance to do something with it, before Peacock returns (eg. additional messages that may arrive after the timer ended, but aren't in the curated list of allowed events).

### DIFF
--- a/components/eventHandler.ts
+++ b/components/eventHandler.ts
@@ -545,9 +545,6 @@ function saveEvents(
             return
         }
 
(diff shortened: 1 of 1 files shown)


RDIL (approved):
No idea how this happened, but thanks for this fix!

PR #24, opened 2022-11-13
fix the order of the h2 escalations

riisikumi:
the h2 escalations are now in the correct order in the menus. based!

Signed-off-by: riisikumi <54016129+riisikumi@users.noreply.github.com>


AnthonyFuller (approved):
LGTM, now do the rest :+1:

Issue #12, opened 2022-11-12
Duplicate tiles of finished challenges at the end of the mission


moonysolari:
Peacock version: 5.4.0
Game: Hitman III (Epic games)

Currently, finished challenges will have duplicate tiles at the end of the mission. This applies to any finished challenges, not only those that were finished during the current run, but also those finished in previous runs.

![image](https://user-images.githubusercontent.com/118079569/201493833-60b75703-4cc4-4df2-bf79-c6f3ddc04ec3.png)

moonysolari:
This is at least partly caused by the incorrect `MissionReward.Challenges` property in the return value of the `missionEnd` function in `components/scoreHandler.ts`. It currently returns all finished challenges of this sublocation. The correct behavior is to only return both global and sublocation-specific challenges that were finished since the last time the `missionEnd` function was called on this sub-location. Note that challenges will accumulate over several sessions if the user deliberately avoids triggering the mission ending page. 

One solution is to categorize the `Extensions.ChallengeProgression` objects by their sublocations, and add a Boolean property for each denoting whether it has previously been returned by the `missionEnd` function (i.e. "ticked" on the mission ending page). 

https://github.com/thepeacockproject/Peacock/blob/fdfe8ff77980829036f380369bcd392ac5a51702/components/scoreHandler.ts#L287-L314

PR #25, opened 2022-11-15
Freelancer and CTT support

AnthonyFuller:
This adds support for Freelancer (Evergreen) and for people connecting with the Closed Technical Test.

To-do:
- [ ] Mission End screens
- [ ] Mastery
- [x] Think about how we're going to deal with allunlockables (EG differs from retail)

Possibly anything else that differs.


static/allunlockables.json
```diff
@@ -17913,6 +17913,29 @@
         "IsDroppable": false,
         "Rarity": "common"
     },
+    {
```
RDIL, line 17916: Potentially breaking.

RDIL (approved):

LennardF1989:
I would propose a squash of all changes to a single commit before merging.

RDIL:
@LennardF1989 we’ll squash it when merging.

AnthonyFuller:
> Most of the stuff that's added are EG only things that won't show up in non-eg anyways.

You'd think that, but this had been added, and it looks like this on non-EG clients:
image

It also looks almost as equally terrible on the normal build.

And yes, there is broken LOCR for the subtype "evergreen" on the suits, that part I was fine with.

grappigegovert:
Ah interesting. Yeah, I didn't think it would do that, that's quite unfortunate

static/allunlockables.json
AnthonyFuller: Resolved in 1f14cf4158d3291103777356b6c0715b545a37d2

moonysolari:
Did you plan to / already fix #12 with the changes on the Mission End screens on your local branch?

PR #26, opened 2022-11-15
Added flags for easily enabling program options

LennardF1989:
This allows users to enable --hmr and --plugin-dev-host from the options.ini.

For developers, this makes things a lot easier as we can just `run-dev` with no additional trickery.
For normal users, this would mean they don't have to mess around with startup arguments.

Tried to set a "standard" for the flag names:
- "development"-prefix if it's only useful during development (eg. it needs dev dependencies)
- "experimental"-prefix meaning it can be removed, is unstable or will change in the future.

### DIFF
--- a/components/flags.ts
+++ b/components/flags.ts
@@ -85,6 +85,14 @@ const defaultFlags: Flags = {
         desc: "Forcibly disable installed mod checks",
         default: false,
     },
+    experimentalHMR: {
+        desc: "[Experimental] Toggle hot reloading of contracts",
+        default: false,
+    },
+    developmentPluginDevHost: {
+        desc: "[Development - Workspace required] Toggle loading of plugins with a .ts/.cts extension inside the /plugins folder",
+        default: false,
+    },
 }
 
 const OLD_FLAGS_FILE = "flags.json5"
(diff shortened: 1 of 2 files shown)


components/flags.ts
```diff
@@ -85,6 +85,14 @@ const defaultFlags: Flags = {
         desc: "Forcibly disable installed mod checks",
         default: false,
     },
+    experimentalHMR: {
+        desc: "[Experimental] Toggle hot reloading of contracts",
+        default: false,
+    },
+    developmentPluginDevHost: {
+        desc: "[Development only] Toggle loading of plugins with a .ts/.cts extension inside the /plugins folder",
```
RDIL, line 93: Worth noting here too that this requires the dev workspace, not just that it's dev only.

RDIL (changes requested):

components/flags.ts
LennardF1989: Changed it to: `[Development - Workspace required] Toggle loading of plugins with a .ts/.cts extension inside the /plugins folder` and force pushed.

RDIL (approved):

PR #27, opened 2022-11-15
Added hot reload functionality for plugins

LennardF1989:
Implementation of my hot reload system for plugins:
- Is backwards compatible with existing plugins
- Contains two test-plugins to show an old situation and a new situation, will only run when test-mode is enabled.

The following changes have been made:
- Exposes an IPlugin interface plugins can implement, this will require a name for the plugin and optional an "unload" method.
- The "unload" method should undo any interceptors or taps that have been applied, it's the plugins responsibility to keep track of that.
- Added "untap" and "removeInterceptor" to hooks to make unloading possible
- Added an API endpoint to reload a plugin based on its name: `/reloadplugin/<pluginname>`
- Added configuration for plugins to lint properly
- Added convenience functions for plugins to check for Development and test-mode

How to test:
- Default committed plugins will just load and do their thing once on startup.
- `/reloadplugin/TestNormalPlugin` will give a 400 response
- `/reloadplugin/TestHotReloadPlugin` will give a 200 response
- Every time you reload the TestHotReloadPlugin, it will always output the following:
```
[Info] GET /reloadplugin/TestHotReloadPlugin
[Debug] [TestHotReloadPlugin] Unloading...
[Info] [TestHotReloadPlugin] Unloaded!
[Info] Plugin 'TestHotReloadPlugin' has been unloaded!
[Info] Plugin 'TestHotReloadPlugin' has been loaded!
[Debug] [TestHotReloadPlugin] Loading...
[Debug] [TestHotReloadPlugin] interceptor tapped!
[Debug] [TestHotReloadPlugin] interceptor called!
[Debug] [TestHotReloadPlugin] hook called!
[Info] [TestHotReloadPlugin] Loaded!
[Debug] Plugin 'TestHotReloadPlugin' potentially supports hot reloading
```
- If you intentionally break the plugin during a hot reload by wiping the "unload" function and immediately returning true, and then reload a few times, the interceptors/taps of previously loaded plugins will also still be called.
- If you change the "return true" into a "return false" and then reload, the unloading will fail and the user is requested to restart the server instead.

Possible extensions/future:
- Create a "scoped" plugin host that can be disposed and undoes all changes it made. However, the current implementation explicitly asks the developer if hot-reloading is supported AND any hacks that were applied outside of the hook system (eg. trampolines on functions) can also be undone.
- Add a route in the web-ui to reload plugins with the press of a button
- Add support for undoing addMission/addEscalation

### DIFF
--- a/plugins/TestHotReload.plugin.ts
+++ b/plugins/TestHotReload.plugin.ts
@@ -1,0 +1,68 @@
+import { log, LogLevel } from "../components/loggingInterop"
+import { Controller, IPlugin } from "../components/controller"
+import { Intercept } from "../components/hooksImpl"
+
+class TestHotReloadPlugin implements IPlugin {
+    private controller: Controller
+
+    public get name() {
+        return "TestHotReloadPlugin"
+    }
+
+    constructor(controller: Controller) {
+        this.controller = controller
+
+        this.init()
+    }
+
+    private init(): void {
+        log(LogLevel.DEBUG, `[${this.name}] Loading...`)
+
+        const interceptor: Intercept<[], void> = {
+            name: `${this.name}_ServerStart_Interceptor`,
+            call: () => {
+                log(LogLevel.DEBUG, `[${this.name}] interceptor called!`)
+            },
+            tap: () => {
+                log(LogLevel.DEBUG, `[${this.name}] interceptor tapped!`)
+            },
+        }
+        this.controller.hooks.serverStart.intercept(interceptor)
+
+        this.controller.hooks.serverStart.tap(
+            `${this.name}_ServerStart_Tap`,
+            () => {
+                log(LogLevel.DEBUG, `[${this.name}] hook called!`)
+            },
+        )
+
+        this.controller.hooks.serverStart.call()
+
+        log(LogLevel.INFO, `[${this.name}] Loaded!`)
+    }
+
+    public unload(): boolean {
+        log(LogLevel.DEBUG, `[${this.name}] Unloading...`)
+
+        let result = this.controller.hooks.serverStart.removeInterceptor(
+            `${this.name}_ServerStart_Interceptor`,
+        )
+        result &&= this.controller.hooks.serverStart.untap(
+            `${this.name}_ServerStart_Tap`,
+        )
+
+        log(LogLevel.INFO, `[${this.name}] Unloaded!`)
+
+        return result
+    }
+}
+
+function initPlugin(controller: Controller): void | IPlugin {
+    if (!controller.runningInTestMode()) {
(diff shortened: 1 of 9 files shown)


LennardF1989:
Force pushed new version that:
- Fixed the linting issue of plugins, while still maintaining support for prettier and such. They will be excluded from building al together.
- Added a toggle to allow running test code and conveniences function to check for development/test-mode in plugins. This effectively allows disabling the test plugins. In absence of a unit test framework, this makes testing things through plugins possible at runtime.

RDIL:
I'm a bit hesitant on this.  We tried implementing hot reloading (albeit on [tapable](https://github.com/webpack/tapable) instead of our own inspired version,) but the number of bugs it caused was very high, to the point where we had to remove it. I think this is the kind of thing that should wait until v6 or v7, as a change this big to the plugin system (even if it's backwards compatible) would almost certainly be both confusing and possibly break.

LennardF1989:
Could you elaborate on the kind of issues that were introduced? 

Mind you this is very much development convenience feature and the time I spend on writing and refactoring it to what it is now already out paid itself back in time saved as I was working on the Evergreen scoring plugin.

We could add it behind an experimental flag to shape out the IPlugin interface (and for hot reloading, only a name and an unload method is required). But I think the beauty is that by having potentially new Plugins implement that interface at the very least, adds future possibilities of extending that interface in a backwards compatible way.

RDIL:
- Plugins that modify configs would cause memory leaks because we would have to store the original config.
- Various object pointers would still point to the plugins before they were reloading, leading to edge cases where the server was still working with the old plugin version, or both at once.

An experimental flag certainly sounds like a good idea, but I think that at least for now, I'd like more people to look at this first, especially @AnthonyFuller, who helped design the first iteration of it.

plugins/TestHotReload.plugin.ts
```diff
@@ -0,0 +1,68 @@
...
+class TestHotReloadPlugin implements IPlugin {
+    private controller: Controller
+
+    public get name() {
+        return "TestHotReloadPlugin"
+    }
+
+    constructor(controller: Controller) {
+        this.controller = controller
+
```
RDIL, line 14:
```suggestion
    public get name() {
        return "TestHotReloadPlugin"
    }

    constructor(private readonly controller: Controller) {
```
Can be simplified

RDIL:
As part of the v6 feature freeze, this is getting pushed back unfortunately.

LennardF1989:
I think I'm gonna retract this PR in light of new features and #219 - The only thing that should be in Peacock is a proper IPlugin interface with defined methods.

PR #29, opened 2022-11-16
Added hook for the MissionEnd response

LennardF1989:
This allows plugins to override the response for the /MissionEnd-call. Returning false will give other plugins a chance, otherwise fallback to the default implementation.

Use-case is implementation a custom scoring and end-screen in a plugin, which is useful to test things before they might be merged into the core files.


AnthonyFuller (approved):
LGTM

RDIL (approved):

PR #30, opened 2022-11-16
Added service for mastery data

LennardF1989:
~~Depends on: #25~~ Added by RDIL, but this is not true. It has portions of which Evergreen can benefit, however.

This PR includes:
- A service to handle/retrieve mastery data
- Mastery data for all locations, including Evergreen (SNUG).
- Added initial support for profile progression on the Player Profile screen as well as when viewing mastery of locations.

This PR does not include:
- Actual progression saving/drops on /MissionEnd, even though I already sorted out mastery drops as part of my custom scoring plugin for Evergreen and the storing of that progression per location in the player profile. I will backport that to general purpose later.

General note:
- This has been tested on Hitman 3 for Epic, and technically the Steam version though the playtest. 
- Most likely will work on Hitman 1 and Hitman 2. Otherwise it would be rather easy to exclude this based on game version.


RDIL:
That was unexpected.

components/candle/masteryService.ts
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        this.masteryData.set(firstMasteryData.CompletionData.Id, masteryData)
+    }
+
+    getMasteryDataForDestination(
+        locationParentId: string,
+        gameVersion: GameVersion,
+        userId: string,
+    ): unknown[] {
```
AnthonyFuller, line 34: These should probably be properly typed.

components/controller.ts
```diff
@@ -831,6 +840,26 @@ export class Controller {
         }
     }
 
+    private async _loadMasteryData(): Promise<void> {
```
AnthonyFuller, line 843: This should probably be combined with `_loadChallenges` into a `_loadResources` function.

contractdata/CARPATHIAN/_CARPATHIAN_MASTERY.json
```diff
@@ -0,0 +1,192 @@
...
+                    "Subtype": "startinglocation",
+                    "GameAsset": null,
+                    "ImageId": "",
+                    "RMTPrice": -1,
+                    "GamePrice": -1,
+                    "IsPurchasable": false,
+                    "IsPublished": true,
+                    "IsDroppable": false,
+                    "Capabilities": [],
+                    "Qualities": {},
```
RDIL, line 45: These are duplicated from `allunlockables` - we don't want to waste this much memory on duplicated data.

packaging/buildTasks.mjs
```diff
@@ -34,32 +34,38 @@ async function readJson(filePath) {
...
             await unlink(join(resources, crp))
         }
     }
+}
+
+async function handleFile(resources, name, contents) {
+    const targetName = createHash("md5").update(name).digest("hex")
+
+    await writeFile(
+        join(resources, `${targetName}.crp`),
```
RDIL, line 51: CRP = challenge resource package, this is a new format, and it doesn't really make sense to have it as the same extension if it stores different data.

AnthonyFuller (changes requested):
All in all a good PR and an especially welcome one! Thanks a lot!

static/EvergreenMasteryDataForLocationTemplate.json
```diff
@@ -0,0 +1,688 @@
+{
+    "template": {
```
RDIL, line 2: For the storage of templates, we want the template to be top-level.

RDIL (changes requested):

components/candle/masteryService.ts
```diff
@@ -0,0 +1,124 @@
...
+            )
+
+            evergreenMasteryDataTemplate.data = {
+                MasteryData: masteryData,
+            }
+
+            return evergreenMasteryDataTemplate
+        }
+
+        return {
```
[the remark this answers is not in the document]
RDIL: Building on this - if it is, the return type doesn't make sense to be called `EvergreenMasteryDataTemplate`.

components/candle/masteryService.ts
```diff
@@ -0,0 +1,124 @@
+import { getConfig } from "components/configSwizzleManager"
```
RDIL, line 1: Consistency - should be `../components/`

packaging/buildTasks.mjs
```diff
@@ -78,9 +84,15 @@ export async function packContractsAndChallenges() {
 
         const json = await readJson(path)
 
-        if (filename.startsWith("_")) {
+        if (filename.endsWith("_CHALLENGES.json")) {
```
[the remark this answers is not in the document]
AnthonyFuller: This may be invalid as it is probably required to detect if the extension should be `crp` or `mrp`.

RDIL (changes requested):

contractdata/AMBROSE/_AMBROSE_MASTERY.json
```diff
@@ -0,0 +1,598 @@
+[
```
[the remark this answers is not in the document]
LennardF1989: The completion data is reset and includes the MaxLevel - this is not always 20, but sometimes 5 and in case of evergreen 25 and in the full release even more.

The reason I have the format as is, is because it significantly simplifies the loading of the data into the proper locations. Otherwise you will either end up:
- Generating this exact output when loading the .csp file with the data (startup overhead, no memory difference)
- Generating this data every time you need it (aka: mapping - no startup overhead, less static memory)

So it's a give or take situation. From a maintainability perspective a smaller revised format is easier, but it comes at a cost.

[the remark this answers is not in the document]
AnthonyFuller: Adding onto this, `TypeLocaKey` can be generated from the unlockable, so here's an updated format:
```json
{
    "Id": "LOCATION_PARENT_ROCKY",
    "Levels": [
        {
            "Level": 2,
            "Unlockable": "AGENCYPICKUP_ROCKY_SMALL_BIGCENTRALTREE",
            "IsLevelMarker": false
        }
    ]
}
```

packaging/buildTasks.mjs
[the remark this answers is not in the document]
LennardF1989: I thought about adding extension support instead, but the easier solution was going with the existing packing logic. 

Two choices:
- Implement multiple extension support, de-duplicate code. Everything could go into a single folder.
- Go for a more generic extension, keep multiple folder logic.

packaging/buildTasks.mjs
```diff
@@ -34,32 +34,38 @@ async function readJson(filePath) {
...
+async function handleFile(resources, name, contents) {
+    const targetName = createHash("md5").update(name).digest("hex")
+
+    await writeFile(
+        join(resources, `${targetName}.crp`),
+        packer.pack(contents),
+    )
+}
+
+export async function packContractsAndChallenges() {
```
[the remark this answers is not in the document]
LennardF1989: Agree, but see above comment. I guess it comes down on how we want to handle packing this data (and potentially more in the future).

contractdata/AMBROSE/_AMBROSE_MASTERY.json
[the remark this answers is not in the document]
AnthonyFuller: > The completion data is reset and includes the MaxLevel
What do you measn by "reset", and "MaxLevel" can be added by a property, just like Id.

I'm sure @RDIL can comment about the generation part, I think having the smaller format is more beneficial as we're not storing the _exact_ same data twice, and it's actually better for maintaining it as if the unlockable changes between updates (i.e. a suit or weapon), we'd have to update the mastery json too while we'd normally just update the one `allunlockables`.

packaging/buildTasks.mjs
LennardF1989: See my comment on Anthony's response to that.

contractdata/CARPATHIAN/_CARPATHIAN_MASTERY.json
LennardF1989: Agree, but see my comment under Anthony's - I think we have to pick a lesser evil here. I'm all for de-deplication though.

packaging/buildTasks.mjs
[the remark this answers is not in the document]
AnthonyFuller: The latter is probably the better of the two options.

components/candle/masteryService.ts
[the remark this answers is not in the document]
LennardF1989: It's currently only possible to see mastery data when in the Safehouse ingame menu. Other location do not do this (yet?). For Evergreen, it has to stay, so I would propose a better name for this then.

contractdata/AMBROSE/_AMBROSE_MASTERY.json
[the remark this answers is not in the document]
RDIL: Unless the duplication has a very good reason for being present, I'd like to avoid it.  Very likely, reducing memory by, say, 10 megabytes, is a better improvement then saving less then say 50 milliseconds.

components/controller.ts
LennardF1989: See comment above. I'm all for going for a generic resource package format. If this were to become loadResources that makes sense. However, from a Single Responsibility point of view while it does load resources, it also does the registration for the specific type of the data it found. I would rather go for a _loadResources-function that does the byte reading, and then two functions (as callbacks?) to handle the actual processing.

components/candle/masteryService.ts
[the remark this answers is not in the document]
AnthonyFuller: I checked this in Retail files, the `MasteryDataForLocation` endpoint is referenced elsewhere irregardless of EG.

contractdata/AMBROSE/_AMBROSE_MASTERY.json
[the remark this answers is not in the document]
AnthonyFuller: It also isn't something that would be ran often so it wouldn't impact much.

components/candle/masteryService.ts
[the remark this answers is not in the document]
RDIL: Fairly sure you can in Sniper Assassin too.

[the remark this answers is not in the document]
AnthonyFuller: It's specifically used when pressed tiles displayed in Career -> Mastery on the main menu, good reuse.

contractdata/AMBROSE/_AMBROSE_MASTERY.json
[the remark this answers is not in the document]
LennardF1989: > What do you mean by "reset", and "MaxLevel" can be added by a property, just like Id.
The MaxLevel determines the maximum level of the mastery without having the scan through the mastery data to grab the highest available level.

But I understand where you guys want to go. Runtime generation of the full data on-demand it is.

LennardF1989:
Added new commits with the requested changes. Purposely didn't squash so you can easily spot the changes.

While I've fixed a performance issue now that everything is generated runtime, I must add it happens quite often: Every time you go into a Destination. So when you are casually browsing to decide the contract to play, you are constantly doing a lot of calculations to build this object with drops. 

If that is acceptable in order to safe some memory, then I'm okay with it.

components/index.ts
```diff
@@ -505,6 +505,7 @@ function startServer(options: { hmr: boolean; pluginDevHost: boolean }): void {
         "contractSessions",
         "userdata",
         "contracts",
+        "plugins",
```
LennardF1989, line 508: This is actually a bug-fix for the hot-reload. I will remove those when I squash after the second round of checks.

contractdata/AMBROSE/_AMBROSE_MASTERY.json
```diff
@@ -0,0 +1,85 @@
+{
+  "Id": "LOCATION_PARENT_ROCKY",
```
AnthonyFuller, line 2: General comment about the mastery data:
If you require max level, you can add it under here e.g.
```json
{
    "Id": "LOCATION_PARENT_ROCKY",
    "MaxLevel": 20,
    "Drops": [...]
}
```

components/contracts/dataGen.ts
```diff
@@ -97,24 +97,34 @@ export function generateCompletionData(
...
+    const completionData = controller.masteryService.getCompletionData(
+        locationId,
+        gameVersion,
+        userId,
+    )
+
+    if (!completionData) {
+        return <CompletionData>{
+            Level: 20,
+            MaxLevel: 20,
```
AnthonyFuller, line 115: See max level comment in the mastery data. Also, if this is default completion data, why is it max level, and if it's meant to be fully completed, why is XpLeft not 0?

components/candle/masteryService.ts
```diff
@@ -0,0 +1,196 @@
...
+        locationId: string,
+        gameVersion: GameVersion,
+        userId: string,
+    ): MasteryDataTemplate {
+        const subLocation: Unlockable = getSubLocationByName(
+            locationId,
+            gameVersion,
+        )
+
+        const evergreenMasteryDataTemplate: unknown =
```
AnthonyFuller, line 53: Rename this variable as it's not EG specific.

AnthonyFuller (changes requested):
This is definitely getting closer to how I envisioned a mastery system. Just a few final things.

components/contracts/dataGen.ts
LennardF1989: This is literally what it was originally (just check the v5 branch). But you are correct, it doesn't make sense.

contractdata/AMBROSE/_AMBROSE_MASTERY.json
LennardF1989: I determined the MaxLevel based on the highest mastery drop, so not needed. They all match up. Only exception would be Evergreen, where there is a Mastery up til 25, but we could technically make it go indefinately (especially since the Safehouse has more unlocks in it post level-25).

LennardF1989:
Since all changes are most likely fully conform agreements, I went ahead and squashed the branch already for a final review.

components/contracts/dataGen.ts
AnthonyFuller: The reason v5 does this is because it doesn't have a progression system so you can't get mastery, it used to be Level 1, but it was changed to allow easy comparison with official server responses when developing the challenge system.

contractdata/AMBROSE/_AMBROSE_MASTERY.json
AnthonyFuller: Adding MaxLevel as a property would reduce the time seeking for the max level, having it default to 20 and an optional property sounds good. Only exceptions are Evergreen and Carpathian.

components/candle/masteryService.ts
```diff
@@ -0,0 +1,198 @@
...
+        }
+
+        const locationData =
+            userProfile.Extensions.progression.Locations[
+                lowerCaseLocationParentId
+            ]
+
+        const maxLevel =
+            masteryData.MaxLevel ||
+            masteryData.Drops.reduce((currentMax, drop) => {
```
AnthonyFuller, line 112: Do you think having this just default to 20 would be more beneficial? As there are only 2 exceptions (Evergreen and Carpathian being 25-100 and 5 respectively).

AnthonyFuller (changes requested):
We're almost there, I can feel it! Just one final thing.

components/candle/masteryService.ts
LennardF1989: Yes, I would go with a default of 20 and the let the JSON tell you if it's supposed to be something else.

components/candle/masteryService.ts
```diff
@@ -0,0 +1,198 @@
...
+                lowerCaseLocationParentId
+            ]
+        ) {
+            userProfile.Extensions.progression.Locations[
+                lowerCaseLocationParentId
+            ] = {
+                Xp: 0,
+                Level: 1,
+            }
+        }
```
RDIL, line 103:
```suggestion
        userProfile.Extensions.progression.Locations[
                lowerCaseLocationParentId
        ] ??= {
            Xp: 0,
            Level: 1,
        }
```

components/candle/masteryService.ts
```diff
@@ -0,0 +1,198 @@
...
+                true,
+            )
+
+        const masteryData = this.getMasteryData(
+            subLocation.Properties.ParentLocation,
+            gameVersion,
+            userId,
+        )
+
+        return <MasteryDataTemplate>{
```
RDIL, line 65: Why is this cast needed?

components/candle/masteryService.ts
```diff
@@ -0,0 +1,198 @@
...
+            return <MasteryDrop>{
+                IsLevelMarker: false,
+                Unlockable: unlockable,
+                Level: drop.Level,
+                IsLocked: drop.Level > completionData.Level,
+                TypeLocaKey: `UI_MENU_PAGE_MASTERY_UNLOCKABLE_NAME_${unlockable.Type}`,
+            }
+        })
+
+        return <MasteryData[]>[
```
RDIL, line 191: Same here.

components/candle/masteryService.ts
```diff
@@ -0,0 +1,198 @@
...
+                log(LogLevel.DEBUG, `No unlockable found for ${drop.Id}`)
+
+                return false
+            }
+
+            return true
+        }).map((drop) => {
+            const unlockable: Unlockable = unlockableMap.get(drop.Id)
+
+            return <MasteryDrop>{
```
RDIL, line 182: Same here.

components/contracts/dataGen.ts
```diff
@@ -97,24 +97,39 @@ export function generateCompletionData(
...
+        userId,
+    )
+
+    if (!completionData) {
+        log(
+            LogLevel.DEBUG,
+            "Could not get CompletionData for location ${locationId}",
+        )
+
+        return <CompletionData>{
```
RDIL, line 118: Same here.

components/candle/masteryService.ts
```diff
@@ -0,0 +1,198 @@
...
+} from "../../components/types/mastery"
+import {
+    CompletionData,
+    GameVersion,
+    Unlockable,
+} from "../../components/types/types"
+import { xpRequiredForLevel } from "../../components/utils"
+
+export class MasteryService {
+    // @ts-expect-error Unused
```
RDIL, line 23: Why have this if unused?

components/contracts/dataGen.ts
AnthonyFuller: This cast is still here!

contractdata/SNUG/_SNUG_MASTERY.json
```diff
@@ -0,0 +1,186 @@
+{
```
AnthonyFuller, line 1: Remove this file, it will be added by the Evergreen branch/PR.

components/contracts/dataGen.ts
```diff
@@ -216,12 +231,14 @@ export function generateUserCentric(
...
  * @param gameChangers The game changers.
  * @param displayOrder The order in which to display the objectives.
+ * @param IsEvergreenSafehouse Is the contract the safehouse?
  * @returns The converted objectives.
  */
 export function mapObjectives(
     objectives: MissionManifestObjective[],
     gameChangers: string[],
     displayOrder: GroupObjectiveDisplayOrderItem[],
+    IsEvergreenSafehouse = false,
```
AnthonyFuller, line 241: How did this end up here?

AnthonyFuller (changes requested):
A few final final things!

components/contracts/dataGen.ts
LennardF1989: Oops, due to the switch-a-roo after moving it away from the evergreen branch :)

AnthonyFuller (approved):
LGTM. Thank you so much for implementing this!

RDIL (approved):

PR #32, opened 2022-11-19
Fixed issue where loading a save would kick you offline in-game

LennardF1989:
Loading a save after restarting the game/server would fail because the startContract of the profileService was not called. Usually newSession does this for you, but since we are loading an existing session...


RDIL (changes requested):
While this is a very important bug fix, I disagree with the implementation.  We should use the existing challenge states, but move them directly onto the session instead.

LennardF1989:
Agreed, it's a hotfix to prevent a crash rather than a complete solution. Not sure how involved moving that over to a contractSession instead would be, but if the answer is "a lot", we might just add this with a //TODO above it and then make a different PR to overhaul it.

PR #34, opened 2022-11-19
Allow restarting Peacock while the game is active

LennardF1989:
Currently if you restart Peacock, you will lose the in-memory UserProfile that got created during the `/oauth/token`-call, causing pretty much everything to topple over and requiring the user to go offline-mode first. On top of that, you will also lose any active contractSession, because they are only ever persisted if you make a save.

This PR fixes the first issue by checking if the unique_name in the JWT token is already loaded into memory on every request. It does this using a cheap check, so it's only a very minor overhead on each request.

The contractSession issue is not really fixed with this PR, but silently averted by creating a newSession on the fly. It's not a pretty solution, but it works.

All in all, this PR allows developers to make changes to Peacock and restart the server without being forced to go to offline-mode, saving valuable development time. For normal end-users, it's not really useful in a local server situation.

That said, while outside the scope of this PR, we should probably consider using a different mechanism to store volatile data, eg. by storing them in a fast NoSQL-like database (Eg. AceBase: https://github.com/appy-one/acebase)

### DIFF
--- a/components/databaseHandler.ts
+++ b/components/databaseHandler.ts
@@ -108,6 +108,33 @@ export function getUserData(
 }
 
 /**
+ * Only attempt to load a user's profile if it hasn't been loaded yet
+ *
+ * @param userId The user's ID.
+ * @param gameVersion The game's version.
+ */
+export async function cheapLoadUserData(
+    userId: string,
+    gameVersion: GameVersion,
+): Promise<void> {
+    if (!userId || !gameVersion) {
+        return
+    }
+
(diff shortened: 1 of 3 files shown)


LennardF1989:
Due to the developer-oriented nature of this PR, we could also just add a developer-flag for the .use() in index.ts, and inside the !session-check in the eventHandler. Then again, it does have valid uses for normal users in that they can reload Peacock while messing with their plugins.

Small-cart:
I agree with you.When i test contracts, i need restart Peacock and have to offline.

PR #35, opened 2022-11-20
Added H2016 escalations

riisikumi:
Added H2016 escalations that were previously available using the Legacy Escalations plugin.


RDIL:
Going to have to say no on this one.  The overhead this adds is significant, and having this as a plugin makes more sense.

I'm open to integrating the ones that did make it into H2 and H3, but this is just a bit too much.

PR #36, opened 2022-11-20
Move challenge contexts to the session

RDIL:
Closes #32


components/candle/challengeService.ts
```diff
@@ -377,11 +365,15 @@ export class ChallengeService extends ChallengeRegistry {
     ): void {
         const writeQueue: PendingProgressionWrite[] = []
 
-        for (const challengeId of Object.keys(
-            this._challengeContexts[sessionId],
-        )) {
+        if (!session.challengeContexts) {
+            log(LogLevel.WARN, "Session does not have challenge contexts.")
+            log(LogLevel.WARN, "Challenges will be disabled!")
```
AnthonyFuller, line 370: When mastery progression exists, it may be wise to change to "Progression will be disabled" as the global challenges should be tracked by the same system as normal challenges.

AnthonyFuller (approved):

LennardF1989:
Appears to have broken saving al together. I'll investigate where it comes from.

PR #37, opened 2022-11-21
Added vitest and configuration to support unit tests

LennardF1989:
Quite some hours (around 8) of configuration, playing with mocks and setting up a handful of useful examples.

The examples cover two situations that are very common in the code-base of Peacock:
- scoreHandler.ts* has a private method that should not be exported, but is interesting enough to actually test.
- oauthToken.ts uses other modules to do work. You could argue this is more of an integration test, but with the help of a few helper methods, you can easily mock these modules**.

\* We might actually just export it, since people will need it if the use the missionEnd hook to make their own scoring, but alas, it's an example on how to test such a method.

\** While jest also has a `jest.mock`, the `jest.spyOn` gives a lot more fine-grained control in that it doesn't mock the full module and proxies the call by default. This is useful if you want to check if it has been called, but still want to execute the actual logic.

That said, all roads lead to Rome in terms of when and how to mock something. 

I'll leave this as a draft for now, but I think it would be beneficial to the quality of Peacock to start minding the test-ability of the code-base as soon as possible. For example: there are a lot of API endpoints that have an inline function. This is very hard to test and would either require exposing the Express app (and then look for the routes we are interested in), or (the better solution) move them to internal functions and expose them the same way as I did for scoreHandler.ts and/or just export them.

Might look into Vitest on another branch and see what it takes to achieve the same.

Tests are setup using the Arrange-Act-Assert pattern.

### DIFF
--- a/tests/helpers/testHelpers.ts
+++ b/tests/helpers/testHelpers.ts
@@ -1,0 +1,49 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import type * as core from "express-serve-static-core"
+import { RequestWithJwt } from "../../components/types/types"
+import { Mock } from "vitest"
+
+export function asMock<T>(value: T): Mock {
+    return value as Mock
+}
+
+export function mockRequestWithJwt(): RequestWithJwt<core.Query, any> {
+    return <RequestWithJwt<core.Query, any>>{}
+}
+
+export function mockResponse(): core.Response {
+    const response = {
+        status: undefined,
+        json: undefined,
+        end: undefined,
+    }
+
+    const mockImplementation = () => {
+        return response
+    }
+
+    response.status = vi.fn().mockImplementation(mockImplementation)
+    response.json = vi.fn()
+    response.end = vi.fn()
+
+    return <core.Response>response
+}
+
+export function getResolvingPromise<T>(value?: T): Promise<T> {
+    return new Promise((resolve) => {
+        resolve(value)
+    })
+}
+
+export function getMockCallArgument<T>(
+    value: any,
+    call: number,
+    argument: number,
+): T {
+    const calls = asMock(value).mock.calls
+
+    expect(calls.length).toBeGreaterThanOrEqual(call)
+    expect(calls[call].length).toBeGreaterThanOrEqual(argument)
+
+    return asMock(value).mock.calls[call][argument] as T
+}
--- a/tests/setup/globalDefines.ts
+++ b/tests/setup/globalDefines.ts
@@ -1,0 +1,7 @@
+Object.assign(globalThis, {
+    PEACOCK_DEV: false,
+    HUMAN_VERSION: "test",
+    REV_IDENT: 1,
+})
+
+export {}
--- a/tests/src/oauthToken.test.ts
+++ b/tests/src/oauthToken.test.ts
@@ -1,0 +1,132 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { UserProfile } from "../../components/types/types"
+import { handleOauthToken } from "../../components/oauthToken"
+import { sign, verify } from "jsonwebtoken"
+import * as databaseHandler from "../../components/databaseHandler"
+import * as platformEntitlements from "../../components/platformEntitlements"
+import axios from "axios"
+
+import {
+    getMockCallArgument,
+    getResolvingPromise,
+    mockRequestWithJwt,
+    mockResponse,
+} from "../helpers/testHelpers"
+
+describe("oauthToken", () => {
+    const pId = "00000000-0000-0000-0000-000000000047"
+
+    const getExternalUserData = vi
+        .spyOn(databaseHandler, "getExternalUserData")
+        .mockResolvedValue("")
+    const loadUserData = vi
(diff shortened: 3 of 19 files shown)


components/scoreHandler.ts
```diff
@@ -679,3 +679,16 @@ export async function missionEnd(
...
+ */
+interface Testing {
+    calculatePlaystyle: typeof calculatePlaystyle
+}
+
+export const _testing: Testing = <Testing>{}
+
+if (PEACOCK_TEST) {
+    _testing.calculatePlaystyle = calculatePlaystyle
+}
```
RDIL, line 694: This feels like a bad idea.

tests/setup/jest.globals.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const g = global as any
+
+g.asMock = asMock
+g.mockRequestWithJwt = mockRequestWithJwt
+g.mockResponse = mockResponse
+g.getResolvingPromise = getResolvingPromise
+g.getMockCallArgument = getMockCallArgument
```
RDIL, line 50: Can't we do this as a module like everything else?

tests/src/jest.test.ts
```diff
@@ -0,0 +1,16 @@
+describe("jest", () => {
+    /**
+     * This test will verify if globals are configured correctly
```
RDIL, line 3: I mean sure, but this feels like testing if a new-born dog has legs.  It's going to be true unless something in the setup process goes horribly wrong.

tests/src/scoreHandler.test.ts
```diff
@@ -0,0 +1,29 @@
...
+        const ratingKill: RatingKill = <RatingKill>{
+            KillClass: "ballistic",
+            KillItemCategory: "pistol",
+        }
+
+        const contractSession = <ContractSession>{
+            kills: new Set<RatingKill>([ratingKill]),
+        }
+
+        const result = _testing.calculatePlaystyle(contractSession)
```
RDIL, line 25: I'd be fine with exporting things like calculatePlaystyle - makes sense.  If we don't want people to use them (outside of tests), we can use `@internal`.

RDIL (changes requested):

tests/setup/globalDefines.ts
```diff
@@ -1,5 +1,5 @@
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-const g = global as any
+const g = globalThis as any
```
LennardF1989, line 2: Unfortunately, this was the only way I could find to get something in the global scope. Using config or exports simply wouldn't work. Same counts for the global testHelpers.

tests/setup/jest.globals.ts
LennardF1989: I moved the functions to a module, but to get them into the global this is the way.

tests/src/jest.test.ts
LennardF1989: Agreed, this test was more simple test-example than anything else. However, it does somewhat validate the integrity of the tests that are run. So a bit of both really.

components/scoreHandler.ts
LennardF1989: Yep, but if you don't want people to complain about removing @internal methods they did end up using, this is the way. They will simply never get a reference when Peacock is run normally. This is probably best compared to the internal keyword found in, for example, C#. Where everything within the code-base is allowed to access it.

LennardF1989:
Much test, very speedy. Ready for review. 

After the initial startup of 2s when using vitest (without --run) or vitest-ui (Vite is the culprit here by the way, not Peacock), re-running/updating the existing tests take only 44ms.

Also recommend adding "yarn test" output is a validation on PRs. Code coverage might also be an interesting statistic.

tests/src/oauthToken.test.ts
```diff
@@ -0,0 +1,112 @@
...
+        expect((accessToken.payload as any).unique_name).toBe(pId)
+    })
+
+    it("epic auth hitman 3", async () => {
+        const request = mockRequestWithJwt()
+        request.body = {
+            grant_type: "external_epic",
+            epic_userid: "0123456789abcdef0123456789abcdef",
+            access_token:
+                "eg1~" +
```
RDIL, line 69: Fairly certain this *only* happens for H2016 epic tokens.

tests/helpers/testHelpers.ts
```diff
@@ -0,0 +1,49 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import type * as core from "express-serve-static-core"
+import { RequestWithJwt } from "components/types/types"
```
RDIL, line 3:
```suggestion
import { RequestWithJwt } from "../../components/types/types"
```

RDIL (changes requested):

LennardF1989:
I really want to get this in ASAP, would have prevented a lot of regression with the score screen...

moonysolari:
I agree. Might make it easier to know whether something is broken.

RDIL (approved):

PR #38, opened 2022-11-22
Fixed performance issue with the configSwizzleManager

LennardF1989:
Don't ask me why, but using "require" instead of "import" shaves off 10 seconds on the startup time, while the resulting configs object is still the same. Pretty much instant startup!

Ran into this as I was trying to get a sort of lazy loading system going (instead of all the imports), since tests suffer from a long startup time due to this module.

### DIFF
--- a/components/configSwizzleManager.ts
+++ b/components/configSwizzleManager.ts
@@ -22,88 +22,6 @@
 import { existsSync, mkdirSync, readFileSync, writeFileSync } from "fs"
 import { join } from "path"
 import { log, LogLevel } from "./loggingInterop"
-
-import Roadmap from "../static/Roadmap.json"
-import StoreData from "../static/StoreData.json"
-import FilterData from "../static/FilterData.json"
-import LocationsData from "../static/LocationsData.json"
-import GameChangerProperties from "../static/GameChangerProperties.json"
-import allunlockables from "../static/allunlockables.json"
-import Destinations from "../static/Destinations.json"
-import connectionConfigTemplate from "../static/config.json"
-import onlineconfig from "../static/onlineconfig.json"
-import privacypolicy from "../static/privacypolicy.json"
-import UserDefault from "../static/UserDefault.json"
-import AgencyPickups from "../static/AgencyPickups.json"
-import Entrances from "../static/Entrances.json"
-import LeaderboardEntriesTemplate from "../static/LeaderboardEntriesTemplate.json"
-import LeaderboardsViewTemplate from "../static/LeaderboardsViewTemplate.json"
-import MissionEndReadyTemplate from "../static/MissionEndReadyTemplate.json"
-import MissionEndNotReadyTemplate from "../static/MissionEndNotReadyTemplate.json"
-import SelectAgencyPickupTemplate from "../static/SelectAgencyPickupTemplate.json"
-import SelectEntranceTemplate from "../static/SelectEntranceTemplate.json"
-import StashpointTemplate from "../static/StashpointTemplate.json"
-import LoadMenuTemplate from "../static/LoadMenuTemplate.json"
-import SaveMenuTemplate from "../static/SaveMenuTemplate.json"
-import Playstyles from "../static/Playstyles.json"
-import HubPageData from "../static/HubPageData.json"
-import DashboardCategoryEscalation from "../static/DashboardCategoryEscalation.json"
-import GlobalChallenges from "../static/GlobalChallenges.json"
-import ContractsTemplate from "../static/ContractsTemplate.json"
-import CreateContractPlanningTemplate from "../static/CreateContractPlanningTemplate.json"
-import CreateContractReturnTemplate from "../static/CreateContractReturnTemplate.json"
-import PlayerProfilePage from "../static/PlayerProfileView.json"
-import Legacyallunlockables from "../static/Legacyallunlockables.json"
-import LegacyGlobalChallenges from "../static/LegacyGlobalChallenges.json"
-import LegacySafehouseTemplate from "../static/LegacySafehouseTemplate.json"
-import LegacyHubTemplate from "../static/LegacyHubTemplate.json"
-import LegacyPlanningTemplate from "../static/LegacyPlanningTemplate.json"
-import LegacySelectAgencyPickupTemplate from "../static/LegacySelectAgencyPickupTemplate.json"
-import LegacySelectEntranceTemplate from "../static/LegacySelectEntranceTemplate.json"
-import LegacyStashpointTemplate from "../static/LegacyStashpointTemplate.json"
-import LegacyUserDefault from "../static/LegacyUserDefault.json"
-import LegacyContractSearchResponseTemplate from "../static/LegacyContractSearchResponseTemplate.json"
-import LegacyFilterData from "../static/LegacyFilterData.json"
-import PlayNextTemplate from "../static/PlayNextTemplate.json"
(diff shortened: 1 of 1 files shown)


AnthonyFuller:
Why?

LennardF1989:
> Why?

As explained, it reduces startup time significantly without any loss of functionality. This is a difference of running a single unit test in about 5 seconds compared to 3ms. In unit tests each file is run in its own context, thus whenever you import a module that imports the configSwizzleManger, you get the overhead it brings. Not only annoying when you are developing tests, the total time spent testing when testing a full suit is baffling. So that's why :p

Tried mocking the file before it loads, but no dice. On top of that, who doesn't like faster startup times?

That said, the another question in this particular piece of code is if all those JSON object have to be loaded AND in memory all the time. But that's a different discussion.

LennardF1989:
# Test-code (pseudo)
```ts
const beforeTime = Date.now()

import Roadmap from "../../static/Roadmap.json"
//or
const Roadmap = require("../../static/Roadmap.json")
//etc...

const configs: Record<string, any> = {
    Roadmap,
    //etc...
}

const keys = Object.keys(configs).map((cfg) => Object.keys(configs[cfg])[0])

console.log(JSON.stringify(keys), `${(Date.now() - beforeTime) / 1000} seconds`)

it("default", () => {
    expect(configs["LegacyHubTemplate"].buttons[0].actiontype).toBe("accept")
})
```

# Top-level import
```
["$datacontext","Categories","0","parents","controller","id","91cf5610-e7ea-4665-9c7b-4b3d5a18a37d","0","0","Versions","Contracts","view","Id","assembly:/_pro/scenes/missions/ancestral/scene_ancestral_vesper.entity","assembly:/_pro/scenes/missions/ancestral/scene_bulldog.entity","controller","controller","metricscontext","buttons","metricscontext","reopen-page","template","0","$datacontext","$if $not $isnull $.data.Item.Data","0","buttons","$setpageargs","title","template","0","0","buttons","$setpageargs","$setpageargs","metricscontext","buttons","metricscontext","Id","0","id","pageisclosable","pageisclosable","f246e3ab-3a35-49a0-a159-615632d7fa89","controller","$if","Categories","0","$setpageargs","parents","onpageopened","reopen-page","id","$if $($arg _isonline)","pageisclosable","$datacontext","4d45b6b0-3bea-4a84-a837-923ec5872fb7","buttons","0","0","Categories","id","parents","0","$datacontext","view","view","$setpageargs","controller","prologue_intermission1","buttons","id","$setpageargs","LOCATION_AUSTRIA","0","assembly:/_pro/scenes/missions/paris/location_parisnoel.brick","0","controller","e47d6995-ba9a-4d13-b40a-84e104b84bb6","0","$datacontext"]
3.235 seconds
```
Time range between 3.1s and 3.5s, 3.2s is the most common.

# Top-level require
```
["$datacontext","Categories","0","parents","controller","id","91cf5610-e7ea-4665-9c7b-4b3d5a18a37d","0","0","Versions","Contracts","view","Id","assembly:/_pro/scenes/missions/ancestral/scene_ancestral_vesper.entity","assembly:/_pro/scenes/missions/ancestral/scene_bulldog.entity","controller","controller","metricscontext","buttons","metricscontext","reopen-page","template","0","$datacontext","$if $not $isnull $.data.Item.Data","0","buttons","$setpageargs","title","template","0","0","buttons","$setpageargs","$setpageargs","metricscontext","buttons","metricscontext","Id","0","id","pageisclosable","pageisclosable","f246e3ab-3a35-49a0-a159-615632d7fa89","controller","$if","Categories","0","$setpageargs","parents","onpageopened","reopen-page","id","$if $($arg _isonline)","pageisclosable","$datacontext","4d45b6b0-3bea-4a84-a837-923ec5872fb7","buttons","0","0","Categories","id","parents","0","$datacontext","view","view","$setpageargs","controller","prologue_intermission1","buttons","id","$setpageargs","LOCATION_AUSTRIA","0","assembly:/_pro/scenes/missions/paris/location_parisnoel.brick","0","controller","e47d6995-ba9a-4d13-b40a-84e104b84bb6","0","$datacontext"] 
0.16 seconds
```
Time range between 0.14s and 0.24s, 0.16s is the most common.

# Inline require (as per this PR)
```
["$datacontext","Categories","0","parents","91cf5610-e7ea-4665-9c7b-4b3d5a18a37d","0","0","Versions","Contracts","view","Id","assembly:/_pro/scenes/missions/ancestral/scene_ancestral_vesper.entity","assembly:/_pro/scenes/missions/ancestral/scene_bulldog.entity","id","controller","controller","controller","metricscontext","buttons","metricscontext","reopen-page","template","0","$datacontext","$if $not $isnull $.data.Item.Data","0","buttons","$setpageargs","title","template","0","0","buttons","$setpageargs","$setpageargs","metricscontext","buttons","metricscontext","Id","id","0","id","pageisclosable","pageisclosable","f246e3ab-3a35-49a0-a159-615632d7fa89","controller","$if","Categories","0","$setpageargs","parents","onpageopened","reopen-page","pageisclosable","$datacontext","4d45b6b0-3bea-4a84-a837-923ec5872fb7","0","0","Categories","id","parents","0","$datacontext","buttons","view","view","controller","$setpageargs","prologue_intermission1","buttons","id","$if $($arg _isonline)","$setpageargs","LOCATION_AUSTRIA","0","assembly:/_pro/scenes/missions/paris/location_parisnoel.brick","0","controller","e47d6995-ba9a-4d13-b40a-84e104b84bb6","0","$datacontext"] 
0.16 seconds
```
Time range between 0.14s and 0.24s, 0.16s is the most common. No difference compared to top-level require.

# Results
All outputs are exactly the same, proving that by the time the `console.log` at the end is executed, all data is actually loaded and in the `configs` constant. `require` is 20x slower than `import`. Still not completely sure why though, as TypeScript compiles the code to CommonJS modules (as per `tsconfig.json`), which in turn also use `require`...

RDIL:
The difference lies with type checking I believe.  But either way, I think this breaks how it’s bundled in production builds.

LennardF1989:
Good point, I'll test that. Otherwise it's back to the drawing board :)

LennardF1989:
@RDIL Can confirm that building chunk0.js, optimizing it and dropping it into the latest release and starting the game through that, everything still works as intended.

This makes sense, as I would expect `dev-start` to not work if it was broken.

LennardF1989:
I think I figured this out, here is the difference.

With require:
```js
var __commonJS = (cb, mod) => function __require() {
  return mod || (0, cb[__getOwnPropNames(cb)[0]])((mod = { exports: {} }).exports, mod), mod.exports;
};

// static/Playstyles.json
var require_Playstyles = __commonJS({
  "static/Playstyles.json"(exports, module2) {
    module2.exports = [
        //JSON
    ];
  }
});

var configs = {
  Playstyles: require_Playstyles()
}
```

With import:
```js
// static/Playstyles.json
var Playstyles_default = [
    //JSON
]

var configs = {
  Playstyles: Playstyles_default
}
```

What's funny is that if I run a test on this code with the largest JSON file in static (LegacyHubTemplate.json at 13MB), the result is actually "shocking":
```
import 13ms
require 97ms
```

I think the reason why we are seeing the slowdown on startup with imports, is because of the large amount. This causes so much data for the NodeJS intepreter on startup (before it even starts executing), that a require beats it time-wise simply because the NodeJS can delay trying to understand the code (maybe AST related)?

LennardF1989:
In light of a different solution, closing this.

PR #40, opened 2022-11-23
Fixed performance issue with JSON files

LennardF1989:
Makes changes to the build process to handle imported JSON files differently. This significantly increases startup time (favorable for unit testing), as well as pretty significantly decreasing the final build size when building (going from 27 mb to just 7 mb).

Changes for speed: The JSON imports will be loaded as strings instead of JavaScript objects.
Changes for build size: The JSON imports will be reserialized without indentation.

Draft because I want to verify if things still work as intended (they should, though).


packaging/build.mjs
```diff
@@ -27,6 +27,23 @@ await packContractsAndChallenges()
...
+        build.onLoad({ filter: /\.json$/ }, async (args) => {
+            const text = await fs.promises.readFile(args.path)
+
+            return {
+                contents: JSON.stringify(JSON.parse(text)),
+                loader: "text",
+            }
+        })
+    },
+}
```
RDIL, line 45: I think this can just use the [`loader` option](https://esbuild.github.io/api/#loader) instead of making this a plugin.

LennardF1989: Nope, that was my first attempt and it will only treat it as text, and not actually remove indentation like I do now. The JSON loader will turn it into an object, so this is the only way to do it and also "compress" the actual resulting string :)

LennardF1989: `let fs = require("fs")` I want to change this to const though. Using an import doesn't work do to the way esbuild works with plugins.

LennardF1989:
Tested in-game:
- by replacing chunk0.js in latest release 
- by dev-start

Startup team significantly reduced. Build size significantly reduced.

RDIL (approved):

PR #41, opened 2022-11-25
Fixed issue with "invalid media type" errors

LennardF1989:
Since the Yarn4 update, that also updated ExpressJS to a new version, it seems like that responses that immediately do .send don't have a media type set (anymore?). This sets the default media type to "text/plain", most likely like it always had before.

Tested in-game (H3 Epic) including full contract, no issues or weird logs anymore.


LennardF1989:
Accidentally included another branch in the first commit, rebased, don't mind it.

RDIL (approved):

PR #42, opened 2022-11-27
Livesplit - Provide a TimeCalc URL for Season/Trilogy speedruns

alex73630:
Generating and printing a TimeCalc link as required for Season/Full game runs from Speedrun.com

This will print a link like this one in the console, ready to be submitted, using Livesplit output.
[https://solderq35.github.io/fg-time-calc/?mode=0&fs3=1&ft2=1&f3t1=1&f4t0=1&d=:&o1=1&fps=&t1=49&c1=Paris&t2=106&c2=Sapienza&t3=94&c3=Marrakesh&t4=79&c4=Bangkok&t5=76&c5=Colorado&t6=45&c6=Hokkaido&t8=74&c8=Sapienza%20reset%201&t9=75&c9=Sapienza%20reset%202&t10=42&c10=Sapienza%20reset%203&t11=42&c11=Sapienza%20reset%204&t12=27&c12=Colorado%20reset%201&t13=38&c13=Hokkaido%20reset%201](https://solderq35.github.io/fg-time-calc/?mode=0&fs3=1&ft2=1&f3t1=1&f4t0=1&d=:&o1=1&fps=&t1=49&c1=Paris&t2=106&c2=Sapienza&t3=94&c3=Marrakesh&t4=79&c4=Bangkok&t5=76&c5=Colorado&t6=45&c6=Hokkaido&t8=74&c8=Sapienza%20reset%201&t9=75&c9=Sapienza%20reset%202&t10=42&c10=Sapienza%20reset%203&t11=42&c11=Sapienza%20reset%204&t12=27&c12=Colorado%20reset%201&t13=38&c13=Hokkaido%20reset%201)

### DIFF
--- a/components/discordRp.ts
+++ b/components/discordRp.ts
@@ -312,11 +312,7 @@ export function scenePathToRpAsset(
         // haven
         case "assembly:/_pro/scenes/missions/opulent/mission_stingray/scene_stingray_basic.entity":
         case "assembly:/_pro/scenes/missions/opulent/mission_stingray/scene_stingray_arcticthyme.entity":
-            return [
-                "havenlastresort",
-                "The Last Resort",
-                "Haven - The Maldives",
-            ]
+            return ["havenlastresort", "The Last Resort", "Haven"]
 
         // miami
         case "assembly:/_pro/scenes/missions/miami/scene_et_sambuca.entity":
@@ -328,11 +324,11 @@ export function scenePathToRpAsset(
 
         // hawkes bay
         case "assembly:/_pro/scenes/missions/sheep/scene_adonis.entity":
-            return ["elusiveadonis", "The Politician", "New Zealand"]
+            return ["elusiveadonis", "The Politician", "Hawke's Bay"]
         case "assembly:/_pro/scenes/missions/sheep/scene_sheep.entity":
-            return ["hawkenightcall", "Nightcall", "New Zealand"]
+            return ["hawkenightcall", "Nightcall", "Hawke's Bay"]
         case "assembly:/_pro/scenes/missions/sheep/scene_opuntia.entity":
-            return ["opuntia", "Opuntia", "New Zealand"]
+            return ["opuntia", "Opuntia", "Hawke's Bay"]
 
         // sgail
         case "assembly:/_pro/scenes/missions/theark/scene_magpie.entity":
--- a/components/livesplit/liveSplitManager.ts
+++ b/components/livesplit/liveSplitManager.ts
@@ -19,9 +19,17 @@
 import { LiveSplitClient, LiveSplitResult } from "./liveSplitClient"
 import { log, LogLevel } from "../loggingInterop"
 import { getAllCampaigns } from "../menus/campaigns"
-import { Campaign, GameVersion, IHit, Seconds, StoryData } from "../types/types"
+import {
+    Campaign,
+    GameVersion,
+    IHit,
+    LiveSplitTimeCalcEntry,
+    Seconds,
+    StoryData,
+} from "../types/types"
 import { getFlag } from "../flags"
 import { controller } from "../controller"
+import { scenePathToRpAsset } from "../discordRp"
 
 export class LiveSplitManager {
     private readonly _liveSplitClient: LiveSplitClient
@@ -36,6 +44,7 @@ export class LiveSplitManager {
     private _currentMissionTotalTime: number
     private _campaignTotalTime: number
     private _completedMissions: string[]
+    private _timeCalcEntries: LiveSplitTimeCalcEntry[]
     private _raceMode: boolean | undefined // gets late-initialized, use _isRaceMode to access
 
     constructor() {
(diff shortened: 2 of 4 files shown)


fu5ha (changes requested):
Very cool idea, I support it! Unfortunately, there's an important corner case that isn't handled yet:

If you "complete" a mission, but have red guns, or otherwise didn't complete your objective, the autosplitter allows you to just restart the level and it will unsplit. The timecalc info tracking should support this case as well. See lines 111-130 for where this is handled for livesplit itself (i.e. where the unsplit call happens). You should be able to just get the latest entry in the timeCalcEntries and set its isComplete to false, if I'm not missing anything.

components/types/types.ts
```diff
@@ -1258,3 +1258,13 @@ export type ContextScopedStorageLocation = "profile" | "hit" | "session"
 export type CPDStore = Record<string, Record<string, string | number | boolean>>
 
 export type ContractProgressionData = Record<string, string | number | boolean>
+
+/**
+ * LiveSplit-related types
+ */
+export interface LiveSplitTimeCalcEntry {
```
LennardF1989, line 1265: Wondering if types.ts is the place for this. Not sure what the policy is in this regard, but this seems specific to LiveSplit and not to types from the game. Shouldn't it go/remain in the livesplit sub-folder then?

components/discordRp.ts
```diff
@@ -312,11 +312,7 @@ export function scenePathToRpAsset(
         // haven
         case "assembly:/_pro/scenes/missions/opulent/mission_stingray/scene_stingray_basic.entity":
         case "assembly:/_pro/scenes/missions/opulent/mission_stingray/scene_stingray_arcticthyme.entity":
-            return [
-                "havenlastresort",
-                "The Last Resort",
-                "Haven - The Maldives",
-            ]
+            return ["havenlastresort", "The Last Resort", "Haven"]
```
LennardF1989, line 315: The changes in this file are technically unrelated to the livesplit integration.

LennardF1989 (commented):
My 2 cents.

components/discordRp.ts
RDIL: They are - he's using this function to get mission names and such from scene paths.

components/types/types.ts
alex73630: I decided to move it to a new `livesplit.ts` types file but I think it's a bit too much for only one interface, as we don't use more of those on this specific module.

Perhaps, maybe the interface could live in the module code itself?
I'd love to get more feedback on this as I don't really know how we should go with this.

Thoughts @RDIL ?

RDIL: I don't really have a specific preference, they can go wherever tbh

components/livesplit/liveSplitManager.ts
```diff
@@ -378,19 +379,23 @@ export class LiveSplitManager {
...
+            computedTime = 1
             this._currentMissionTotalTime += 1
             this._campaignTotalTime += 1
         } else {
             // important to always floor before adding time
             this._currentMissionTotalTime += Math.floor(time)
             this._campaignTotalTime += Math.floor(time)
         }
+        return computedTime
     }
```
fu5ha, line 399: Added return type and simplified the logic a bit. Also, fixed the fact that the returned time wasn't floored
```ts
    private _addMissionTime(time: Seconds): Seconds {
        let computedTime = Math.floor(time)
        
        // always add at least minimum, which is usually 0 except on cutscenes where
        // you can gain an advantage by restarting in cs (bangkok, sgail specific starts)
        if (time <= this._resetMinimum) {
            computedTime = this._resetMinimum
        } else if (time > 0 && time <= 1) {
            // if in game time is between 0 and 1, add full second
            computedTime = 1
        }

        this._currentMissionTotalTime += computedTime
        this._campaignTotalTime += computedTime
        return computedTime
    }
```

components/livesplit/liveSplitManager.ts
```diff
@@ -426,10 +431,21 @@ export class LiveSplitManager {
...
-    private _generateTimeCalcLink() {
-        const url = new URL(
-            "https://solderq35.github.io/fg-time-calc/?mode=0&fs3=1&ft2=1&f3t1=1&f4t0=1&d=:&o1=1&fps=",
+    private _unsplitLastTimeCalcEntry() {
+        const entry = this._timeCalcEntries.pop()
+        entry.isCompleted = false
+        log(
+            LogLevel.DEBUG,
+            `Unsplitted TimeCalc entry: ${JSON.stringify(entry)}`,
         )
```
fu5ha, line 440: probably get rid of this log before merging

fu5ha (changes requested):
I didn't really look too hard at the link generation code, but other than that looks good other than comments :)

RDIL (approved):
Link gen + discord RP changes LGTM, but Fusha's feedback still remains.

components/livesplit/liveSplitManager.ts
alex73630: Removed!

alex73630:
I updated the code following @fu5ha recommendations and now waiting on approval.

Thanks everyone for the feedback!

components/types/types.ts
alex73630: I'll leave it like this then. We'll probably discuss this later if we have a need to re-organize typings for whatever reason.

fu5ha (approved):

PR #49, opened 2022-12-01
Translated Text in Dynamic Ressources into German

Cookie303728:
Since i always play the game a in german, i was a little bit (only a little bit) annoyed by the english and german mix. Here are my 2 cents contributed to peacock. sadly i couldnt modify the text in most Escelation Bricks since they are located in the escelation file instead of the dynamic ressources file.


Kakiking (approved):
LGTM, didn't find any translation errors

PR #50, opened 2022-12-02
Reworking of challenge system

RDIL:
Talked about in https://github.com/thepeacockproject/Peacock/discussions/45

Just opening for initial thoughts - haven't even tested it yet


components/candle/challengeService.ts
```diff
@@ -811,51 +737,49 @@ export class ChallengeService extends ChallengeRegistry {
...
-        let contract
+        let contract: MissionManifest | null
+
         // TODO: Properly get escalation groups for this
         if (challenge.Type === "contract") {
             contract = this.controller.resolveContract(
                 challenge.InclusionData?.ContractIds?.[0] || "",
             )
 
             // This is so we can remove unused data and make it more like official - AF
```
AnthonyFuller, line 748: This should be investigated as to see whether the format matches official, I have a strong suspicion it many not 100% match. (I remember half arsing the format as it was a mess)

AnthonyFuller (approved):
Logic all LGTM, untested within the game.

PR #51, opened 2022-12-03
Fix documentation for `addClientSideModDependency` and add new `modIsInstalled` function with more sensible naming and semantics

atampy25:
The `addClientSideModDependency` function used to do what it says; it would check if the mod was installed and abort the plugin loading if it wasn't. At some point, the semantics changed to supposedly "returning whether the mod is available", but even that has been broken for a long time, as it actually returns whether the mod is UNavailable. This has had a knock-on effect to the `overrideFrameworkChecks` flag, which is meant to make it so that checks always succeed; however, because a check succeeding means returning false in this broken function, the flag being set actually makes all plugins with mod dependencies abort initialisation.

This PR deprecates the old function and updates its documentation to describe the actual behaviour of the function. It also introduces a new function with a more sensible name that describes its more sensible semantics.

### DIFF
--- a/components/controller.ts
+++ b/components/controller.ts
@@ -432,7 +432,7 @@ export class Controller {
      * Returns whether a mod is UNAVAILABLE.
      *
      * @param modId The mod's ID.
-     * @returns If the mod is unavailable. You should probably abort initialisation if true is returned. Also returns true if the `overrideFrameworkChecks` flag is set.
+     * @returns If the mod is unavailable. You should probably abort initialization if true is returned. Also returns true if the `overrideFrameworkChecks` flag is set.
      * @deprecated since v5.5.0
      */
     public addClientSideModDependency(modId: string): boolean {
--- a/components/controller.ts
+++ b/components/controller.ts
@@ -427,15 +427,32 @@ export class Controller {
     }
 
     /**
-     * Adds a dependency on a client side mod through the Simple Mod Framework.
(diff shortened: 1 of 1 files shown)


components/controller.ts
```diff
@@ -427,11 +427,13 @@ export class Controller {
     }
 
     /**
-     * Adds a dependency on a client side mod through the Simple Mod Framework.
-     * See the cookbook for a usage example!
+     * You should use `modIsInstalled` instead!
```
RDIL, line 430:
```suggestion
     * You should use {@link modIsInstalled} instead!
```

components/controller.ts
```diff
@@ -427,11 +427,13 @@ export class Controller {
...
+     *
+     * Returns whether a mod is UNAVAILABLE.
      *
      * @param modId The mod's ID.
-     * @returns If the mod is available. You will probably want to abort plugin initialization if false is returned.
+     * @returns If the mod is unavailable. You should probably abort initialisation if true is returned. Also returns true if the `overrideFrameworkChecks` flag is set.
```
RDIL, line 435:
```suggestion
     * @returns If the mod is unavailable. You should probably abort initialization if true is returned. Also returns true if the `overrideFrameworkChecks` flag is set.
```
All other docstrings use American English, so this makes sense for consistency's sake.

RDIL (approved):
LGTM

PR #54, opened 2022-12-06
add contract mode in Romania and freeform training

WINNODD:
minor fixs for the tile/background image in snow festival/freeform training

### DIFF
--- a/contractdata/CARPATHIAN/UNTOUCHABLECONTRACTCREATION.json
+++ b/contractdata/CARPATHIAN/UNTOUCHABLECONTRACTCREATION.json
@@ -1,0 +1,36 @@
+{
+    "Data": {
+        "Objectives": [],
+        "Bricks": [],
+        "GameChangers": [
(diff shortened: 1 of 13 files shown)


RDIL (changes requested):
Freeform training sounds fine, but Romania is not a map suitable for contracts.

WINNODD:
I'm aware Romania is no common sandbox like other locations but yet it has every other essential element for contract mode. The reason I made this is to give people a chance to explore this map more than get in and kill the constant.
Just saying we don't have be as thoughtful as IO on such matter. In the end it's just peacock exclusive content.

RDIL:
I understand your reasoning on the matter, but I think this is just not really something we want to have as a first-class feature.  A plugin can do this, and would probably be a better use of the API.

RDIL:
The issue is that the scene file is not designed to support contracts.  It's already very very coupled to the main mission.

RDIL:
Closing due to lack of activity.  Feel free to open a new PR if you are interested in pushing this forward.

PR #55, opened 2022-12-12
Implement support for SMF server replacement compatibility

atampy25:
This is currently just so that Cirrus will build it for me, since Peacock fails to build with a series of bizarre errors whenever I try


atampy25:
LGTM

components/controller.ts
```diff
@@ -479,6 +479,65 @@ export class Controller {
...
+                parse(readFileSync(modFrameworkDataPath!).toString()) as {
+                    lastServerSideStates: {
+                        unlockables?: Unlockable[]
+                        contracts?: {
+                            [k: string]: MissionManifest
+                        }
+                        blobs?: Record<string, string>
+                    }
+                }
+            ).lastServerSideStates
```
AnthonyFuller, line 498: You might as well type this properly, a type may already exist for it as we read it for the deployed mods. If it does not, it should be refactored to use a type.

AnthonyFuller (changes requested):
Other than the comment, it LGTM.

components/controller.ts
AnthonyFuller: Then see the latter part of my comment.

AnthonyFuller (approved):
LGTM

PR #61, opened 2022-12-14
Escalation picker fixes and improvements

moonysolari:
This PR addresses several issues of the escalation picker:
- Improved the error handling by allowing the frontend to show errors messages in the console and push error alerts to the user.
- Added more error types and corresponding error messages.
- Prevents the level shown on the picker to exceed the maximum level of the escalation, or to go below 1. An alert is shown when the user attempts to do either of these.
- Fixed the id of The Baskerville Barney in `static/EscalationCodenames.json`. It previously had the same id as the Dartmoor Garden Show.
- Alerts the user upon attempts to set the level for unregistered escalations, likely due to the user not having the Legacy Escalations plugin.

Besides these, it also:
- Fixes the error message shown when `PEACOCK_DEV` is set to true and the user attempts to visit `localhost`. 

It does not add peacock-exclusive escalations.

### DIFF
--- a/components/index.ts
+++ b/components/index.ts
@@ -127,7 +127,8 @@ app.use("/_wf", webFeaturesRouter)
 
 app.get("/", (req: Request, res) => {
     if (PEACOCK_DEV) {
-        res.send("<html lang=\"en\">PEACOCKDEV active, please run \"npm run webui start\" to start the web UI on port 3000 and access it there.</html>")
+        res.contentType("text/html")
+        res.send("<html> PEACOCKDEV active, please run \"npm run webui start\" to start the web UI on port 3000 and access it there. </html>")
     } 
     else {
         const data = readFileSync("webui/dist/index.html").toString()
--- a/components/index.ts
+++ b/components/index.ts
@@ -128,9 +128,10 @@ app.use("/_wf", webFeaturesRouter)
 app.get("/", (req: Request, res) => {
     if (PEACOCK_DEV) {
         res.contentType("text/html")
-        res.send("<html> PEACOCK_DEV active, please run \"npm run webui start\" to start the web UI on port 3000 and access it there. </html>")
-    } 
-    else {
+        res.send(
+            '<html> PEACOCK_DEV active, please run "npm run webui start" to start the web UI on port 3000 and access it there. </html>',
+        )
+    } else {
         const data = readFileSync("webui/dist/index.html").toString()
 
         res.contentType("text/html")
--- a/components/index.ts
+++ b/components/index.ts
@@ -127,8 +127,9 @@ app.use("/_wf", webFeaturesRouter)
 
 app.get("/", (req: Request, res) => {
     if (PEACOCK_DEV) {
-        res.send("dev active, you need to access the UI from port 3000")
-    } else {
+        res.send("<html lang=\"en\">PEACOCKDEV active, please run \"npm run webui start\" to start the web UI on port 3000 and access it there.</html>")
+    } 
+    else {
         const data = readFileSync("webui/dist/index.html").toString()
 
         res.contentType("text/html")
--- a/components/index.ts
+++ b/components/index.ts
@@ -129,7 +129,7 @@ app.get("/", (req: Request, res) => {
     if (PEACOCK_DEV) {
         res.contentType("text/html")
         res.send(
-            '<html> PEACOCK_DEV active, please run "npm run webui start" to start the web UI on port 3000 and access it there. </html>',
+            '<html> PEACOCK_DEV active, please run "yarn webui start" to start the web UI on port 3000 and access it there. </html>',
         )
     } else {
         const data = readFileSync("webui/dist/index.html").toString()
--- a/components/index.ts
+++ b/components/index.ts
@@ -128,7 +128,7 @@ app.use("/_wf", webFeaturesRouter)
 app.get("/", (req: Request, res) => {
     if (PEACOCK_DEV) {
         res.contentType("text/html")
-        res.send("<html> PEACOCKDEV active, please run \"npm run webui start\" to start the web UI on port 3000 and access it there. </html>")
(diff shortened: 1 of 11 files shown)


moonysolari:
Side note: Currently there are two files where escalation information resides: `static/EscalationCodenames.json` and `components/contracts/escalationMappings`. While I see the reason for having two files, it brings several inconveniences:
- It is more error-prone (the id of The Baskerville Barney).
- Efforts have to be made to update one when another is changed, hence the need to "add" peacock-exclusive escalations to the picker. 

It is high time that this setup be reconsidered, now that the escalation system is under rework.

components/index.ts
```diff
@@ -127,7 +127,10 @@ app.use("/_wf", webFeaturesRouter)
 
 app.get("/", (req: Request, res) => {
     if (PEACOCK_DEV) {
-        res.send("dev active, you need to access the UI from port 3000")
+        res.contentType("text/html")
+        res.send(
+            '<html> PEACOCK_DEV active, please run "npm run webui start" to start the web UI on port 3000 and access it there. </html>',
```
RDIL, line 132: We don't use npm.

components/webFeatures.ts
```diff
@@ -115,24 +125,51 @@ webFeaturesRouter.get(
...
+            isNaN(parseInt(req.query.level)) ||
+            parseInt(req.query.level) <= 0
+        ) {
+            formErrorMessage(res, "The level must be a positive integer.")
             return
         }
 
-        if (!req.query.id && !uuidRegex.test(req.query.id)) {
-            res.json({ error: "bad id" })
+        console.log("Setting level to " + req.query.level)
```
RDIL, line 143: It's better to use our logging solution because it's all nicely formatted.

RDIL (changes requested):

RDIL:
> If a plugin adds new escalations, they will not show up on the picker.
If a plugin explicitly adds an escalation to the list, then it will.

components/index.ts
moonysolari: Changed to yarn.

components/webFeatures.ts
moonysolari: Changed to formatted logging.

components/index.ts
```diff
@@ -127,7 +127,10 @@ app.use("/_wf", webFeaturesRouter)
 
 app.get("/", (req: Request, res) => {
     if (PEACOCK_DEV) {
-        res.send("dev active, you need to access the UI from port 3000")
+        res.contentType("text/html")
+        res.send(
+            '<html> PEACOCK_DEV active, please run "yarn webui start" to start the web UI on port 3000 and access it there. </html>',
```
RDIL, line 132: Why is this now wrapped with an HTML tag?

components/webFeatures.ts
```diff
@@ -38,7 +41,12 @@ if (PEACOCK_DEV) {
         next()
     })
 }
-
+function formErrorMessage(res: Response, message: string): void {
+    res.json({
+        success: false,
+        error: message,
+    })
+}
```
RDIL, line 49:
```suggestion
}

```

components/webFeatures.ts
```diff
@@ -115,24 +126,56 @@ webFeaturesRouter.get(
...
         } catch (e) {
-            res.json({ error: "failed to load user data" })
+            formErrorMessage(res, "Failed to load user data.")
             return
         }
-
+        if (controller.escalationMappings[req.query.id] === undefined) {
+            formErrorMessage(res, "Unknown escalation.")
+            return
+        }
```
RDIL, line 161:
```suggestion
        }

```

components/webFeatures.ts
```diff
@@ -115,24 +126,56 @@ webFeaturesRouter.get(
...
+        if (
+            getLevelCount(controller.escalationMappings[req.query.id]) <
+            parseInt(req.query.level)
```
RDIL, line 164:
```suggestion
            parseInt(req.query.level, 10)
```
Just in case any escalations come along that use double digits, need the proper radix.

components/webFeatures.ts
```diff
@@ -115,24 +126,56 @@ webFeaturesRouter.get(
...
+        ) {
+            formErrorMessage(
+                res,
+                "Cannot exceed the maximum level for this escalation!",
+            )
+            return
+        }
```
RDIL, line 171:
```suggestion
        }

```

components/webFeatures.ts
```diff
@@ -115,24 +126,56 @@ webFeaturesRouter.get(
...
+        log(
+            LogLevel.INFO,
+            "Setting the level of escalation " +
+                req.query.id +
+                " to " +
+                req.query.level,
```
RDIL, line 177:
```suggestion
            `Setting the level of escalation ${req.query.id} to ${req.query.level}`,
```

components/webFeatures.ts
```diff
@@ -50,7 +58,7 @@ webFeaturesRouter.get(
             !req.query.gv ||
             !["h1", "h2", "h3"].includes(req.query.gv ?? null)
         ) {
-            res.json({ error: "bad gv" })
+            res.json([])
```
RDIL, line 61: Why return an empty list here when using that error helper everywhere else?

RDIL (changes requested):

components/index.ts
moonysolari: If I remove the `res.contentType("text/html")` line, I get `TypeError: invalid media type at ServerResponse.send`. The `<html>` flags can be deleted alright. I just wanted to highlight that html is what being sent.

components/webFeatures.ts
moonysolari: Partly because it is supposed to return a list here, but also because `gv` can be `h0` here. Another workaround is add `h0` as a potential value for a `GameVersion`, return an empty list if the `gv` is `h0`, then use that error helper if the `gv` is any other value.

RDIL (approved):

PR #62, opened 2022-12-15
Merging `static/EscalationCodenames.json` and `components/contracts/escalationMappings.ts`

moonysolari:
Building on the [note](https://github.com/thepeacockproject/Peacock/pull/61#issuecomment-1350585839) in #61, I have figured out a way that `EscalationCodenames.json` can be merged into `escalationMappings.ts` in this PR:
- The `EscalationGroup` type will have `number | string` as the type for its keys to accommodate `string`s as keys.
- An entry in `escalationMappings.ts` will contain the `name` and `codename` of an escalation.
- Instead of getting the `EscalationCodenames` config from the `configSwizzleManager`, the `webFeatures` will just pass the `escalationMappings` to the frontend.
- The `CodenameMeta` type is replaced by the `EscalationGroup` type, which is used in both the backend and the frontend.
- The frontend will iterate through the `escalationMappings` and render the picker for each escalation it finds.

The `escalationMappings.ts` is only for demonstration purposes in this PR, hence it only has the `name` and `codename` of three escalations. More will be added if this change is agreed upon.

As for escalation plugins, the game itself will still be compatible with them, but the picker will not be able to find the names of the escalations. This is because the plugins currently do not pass the names of the escalations when calling `addEscalation`. Indeed, the escalation names of all the legacy escalations to show on the UI were managed by Peacock itself, which seems rather illogical considered they are only supported with an extra plugin. 

 There are a couple of ways to fix this:
1. Add a new parameter `name` to `addEscalation` to allow plugins to pass the name of the escalation, and require all escalation plugin writers to update their plugins using this new interface. Using an old version of the plugin will result in an error when starting the server.
2. Add `name` as an optional parameter. Ask all escalation plugin writers to update their plugins using this new interface. Using an old version of the plugin will not result in an error, but if `name` is not passed, then the picker will show the escalation id on the web UI instead. 
3. The same as option 2, but maintain a dict that matches some of the known escalation IDs to their names. If `name` is not passed, look for it first in the dict. If not found, then the picker will show the escalation id on the web UI instead.

I personally prefer option 2. It makes more sense to make it the responsibility of the plugin to pass the names of its escalations.

Another note: it seems that the `codename` of an escalation is currently only used as the `key` in the web UI, in `EscalationLevelPicker.tsx`, but the `id` of the escalation can also do this job. I think it is worth considering to use the `id` here because I see no reason to use an extra string for this.

### DIFF
--- a/components/configSwizzleManager.ts
+++ b/components/configSwizzleManager.ts
@@ -99,7 +99,6 @@ import MasteryUnlockablesTemplate from "../static/MasteryUnlockablesTemplate.jso
 import SniperLoadouts from "../static/SniperLoadouts.json"
 import Scpcallunlockables from "../static/Scpcallunlockables.json"
 import DiscordRichAssetsForBricks from "../static/DiscordRichAssetsForBricks.json"
-import EscalationCodenames from "../static/EscalationCodenames.json"
 import scoreoverviewtemplate from "../static/scoreoverviewtemplate.json"
 import PeacockGameChangerProperties from "../static/PeacockGameChangerProperties.json"
 import MultiplayerPresets from "../static/MultiplayerPresets.json"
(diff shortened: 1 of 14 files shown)


RDIL:
Superseded by #28

PR #68, opened 2023-01-06
Added Mission Story Progression

moonysolari:
## Features: 
- Peacock now no longer treats any mission stories as completed by default. Instead, it tracks the completion of the mission stories in game and writes the repository ID of any completed mission stories to the user data. Deleting the `userdata` folder will reset the mission story progression.
- The number of total mission stories and completed mission stories now show on the Destinations page, as well as on the page for each location.
## Fix:
- Club 27 was previously missing the Mission Stories tab on its planning page. It is added back.
## Breaking change:
- `static/Destinations.json` was deleted, its information merged into `static/LocationsData.json`. 

## Note: 
The game itself also tracks completed mission stories and send `PersistentBool`s to the server via `GamePersistentDataService/SaveData` to communicate this information. The `PersistentBool`s are in the following format:
``` json
{
    "Polarbear_opp_Jet": true,
    "Octopus_opp_Lunch": true,
    "Showstopper_opp_Cocktail": true,
    "Copperhead_OPP_Pyro": true,
    "GECKO_OPP_GREY01": true
}
```
The implementation in this PR ignores the `PersistentBool`s sent by the game and instead tracks the mission stories on the server side for the following reasons:
- Storing the completion data on the server side allows for resetting the mission story progression.
- It is easier to get the location id of the mission story this way. The chaotic naming convention of the `PersistentBool`s make it more challenging.
- The `PersistentBool`s contains all the mission stories ever completed in the game on **any** server, meaning it will also report mission stories that were only completed on official servers but not Peacock.
- I don't have all the names of the mission-story-related `PersistentBool`s.


AnthonyFuller:
I've not properly looked at this yet, but from a glance. You do delete _the import_ of Destinations.json, but you never actually deleted the file.

moonysolari:
> I've not properly looked at this yet, but from a glance. You do delete _the import_ of Destinations.json, but you never actually deleted the file.

You are correct. I added a commit to actually delete it.

components/eventHandler.ts
```diff
@@ -719,6 +724,22 @@ function saveEvents(
...
+                break
+            case "ContractFailed":
+                session.timerEnd = event.Timestamp
+                contractFailed(event, session)
+                break
+            case "OpportunityEvents": {
+                const val = (<OpportunityEventsC2SEvent>event).Value
+                const opportunities = userData.Extensions.opportunityprogression
+                if (val.Event === "Completed") {
+                    opportunities[val.RepositoryId] = true
```
RDIL, line 738: We'll need to call `writeUserData` after modifying this so it's marked as dirty.

static/LocationsData.json
```diff
@@ -2,6 +2,7 @@
     "parents": {
         "LOCATION_PARENT_ICA_FACILITY": {
             "Id": "LOCATION_PARENT_ICA_FACILITY",
+            "Opportunities": 4,
```
RDIL, line 5: Seems dangerous to hard-code this, perhaps we should check all contracts present on that location?

RDIL (changes requested):

static/LocationsData.json
moonysolari: This is the data merged from `static/Destinations.json`, and I think we are safe enough to use it. I haven't seen any mods that attempt at adding mission stories, and I doubt we will get new missions from IOI that add mission stories (even Ambrose does not have any).

RDIL (approved):

PR #71, opened 2023-01-07
Fix image and text details on challenges page

### LINKED ISSUE #66: Image and text for empty categories are missing when filters are applied on challenge pages
**Describe the bug**
When a filter is applied on a challenge page (e.g. SHOW COMPLETED), no image or text is shown for categories that have no challenges to show.

**To reproduce**
- Delete the `userdata` folder.
- Launch the game with Peacock..
- Go to the challenges page of any mission/destination with challenges implemented.
- Change the filter to SHOW COMPLETED.

**Additional context**
image

moonysolari:
- Fixes #66 
- Removes duplicate implementation in `challengeService.ts` as stated in #45. 
- Moves the "Contract Not Found warning" to a more reasonable place, the `resolveContract` function in `components/controller.ts`


RDIL (approved):
LGTM - want a second look from either Tony or Tampy first though.

AnthonyFuller (approved):
Can't see anything logically wrong with it, so LGTM. Thanks for all your recent contributions!

Also, I assume this makes the data we return to the game more like official's? Because I have remarked in the past that it's something we should probably do.

moonysolari:
> Also, I assume this makes the data we return to the game more like official's? Because I have remarked in the past that it's something we should probably do.

It does indeed. I'd say this one is 98% similar to the official responses.

Issue #72, opened 2023-01-08
Wrong token and sessionID saved when game saves


moonysolari:
### Describe the bug
When `/authentication/api/userchannel/ProfileService/UpdateUserSaveFileTable` is accessed, the client sends not just one, but an entire list of metadata of multiple saves. This is currently handled by always taking the **last** item from the list and using its token and sessionID to save the current session. However, this is not always correct, as the list is not sorted with respect to the timestamp.

### To reproduce
- Add `log(LogLevel.DEBUG, JSON.stringify(req.body.clientSaveFileList))` at the start of the function that handles the request (attached below). 
- Compile and run Peacock. Start the game with it.
- Start any mission with saves enabled.
- Try saving the game at different save slots and observe the output. 

### Additional context
Peacock version v5.6.2.

A possible solution is to sort the list with respect to the timestamp and use the token and sessionID of the latest save.

https://github.com/thepeacockproject/Peacock/blob/d95553bd335a3483bada56dd68cce6f675d479c1/components/profileHandler.ts#L630-L659

**Sample output in chronological order (tested on The Finishing Line):**
Saving in save slot 1 caused the client to send the following SaveFile list
``` json
[
   {
      "ContractSessionId":"598676657372700-0b3af383-2f2c-4a2a-a13e-7b687dd37a97",
      "TimeStamp":1673215743,
      "Value":{
         "Name":"QuickSave0",
         "LastEventToken":"598708361135300"
      }
   }
]
```
After that, saving in save slot 2 caused the client to send the following SaveFile list
``` json
[
   {
      "ContractSessionId":"598676657372700-0b3af383-2f2c-4a2a-a13e-7b687dd37a97",
      "TimeStamp":1673215743,
      "Value":{
         "Name":"QuickSave0",
         "LastEventToken":"598708361135300"
      }
   },
   {
      "ContractSessionId":"598676657372700-0b3af383-2f2c-4a2a-a13e-7b687dd37a97",
      "TimeStamp":1673215837,
      "Value":{
         "Name":"QuickSave1",
         "LastEventToken":"598800038190100"
      }
   }
]
```
After that, saving in save slot 3 caused the client to send the following SaveFile list
``` json
[
   {
      "ContractSessionId":"598676657372700-0b3af383-2f2c-4a2a-a13e-7b687dd37a97",
      "TimeStamp":1673215743,
      "Value":{
         "Name":"QuickSave0",
         "LastEventToken":"598708361135300"
      }
   },
[...]

PR #73, opened 2023-01-09
New saving/loading behaviors to fix bugs related to saving

moonysolari:
Tested with casual gameplay, both before and after restarting the game and Peacock.
### New saving/loading behaviors
- When saving,
  - The incoming data is checked against several criteria to reject unwanted requests.
  -  Information about the saved sessions and the slot they correspond to are stored in a new field `Saves` in user profiles.
  - The contract session gets stored to a file in `contractSessions`. The filename is in the format of `slot_token_sessionID.json`.
- When a save slot gets overwritten,
  - The `Saves` field of the user is updated with new information.
  - The file that keeps track of the overwritten session is deleted.
  - A new file is created for the newly stored session.
- When loading,
  - Peacock will load any file with the given token and sessionID in the filename, ignoring the slot name.
### Other changes
- The list of `SaveFile`s received from `/authentication/api/userchannel/ProfileService/UpdateUserSaveFileTable` is correctly processed, such that the one with the latest timestamp is used instead of the last one in the list. 
- The naming convention of the files in `contractSessions` is changed. They now have the slot name prepended.
- Several variable were renamed in `components/databaseHandler.ts` to avoid confusion. There were previously named `sessionid`s but none of them were IDs of a session.
- `saveSession` and `loadSession` were moved to `components/profileHandler.ts`. 
- Removed unused property `PeacockCChild` in user profiles.
- Added some debug info for related functions.
### Issues likely fixed **(need testing)**
- #72 .
- Any problem with loading an autosave, as their sessions were completely unsaved previously.
- Save/load exploit to spam leaderboards with impossible time records.
- Getting incorrect scoring and/or rating after loading a Peacock save.
- Other issues related to saving/loading.


components/databaseHandler.ts
```diff
@@ -216,33 +217,48 @@ export async function writeExternalUserData(
...
-    sessionId: string,
+    identifier: string,
 ): Promise<ContractSession> {
+    const files = await readdir("contractSessions")
+    const filtered = files.filter((fn) => fn.endsWith(`_${identifier}.json`))
+    if (filtered.length === 0) {
+        throw Error(`No session saved with identifier ${identifier}`)
+    }
+    // The filtered files have the same identifier, they are just stored at different slots
+    // So we can read any of them and it will be the same.
```
RDIL, line 232:
```suggestion
    const filtered = files.filter((fn) => fn.endsWith(`_${identifier}.json`))

    if (filtered.length === 0) {
        throw new Error(`No session saved with identifier ${identifier}`)
    }

    // The filtered files have the same identifier, they are just stored at different slots
    // So we can read any of them and it will be the same.
```

components/eventHandler.ts
```diff
@@ -682,14 +677,23 @@ function saveEvents(
             case "IntroCutEnd":
                 if (!session.timerStart) {
                     session.timerStart = event.Timestamp
+                    log(
+                        LogLevel.DEBUG,
+                        `Mission started at : ${session.timerStart}`,
```
RDIL, line 682:
```suggestion
                        `Mission started at: ${session.timerStart}`,
```

components/eventHandler.ts
```diff
@@ -682,14 +677,23 @@ function saveEvents(
...
+                    )
                 }
                 break
             case "exit_gate":
                 session.timerEnd = event.Timestamp
+                log(LogLevel.DEBUG, `Mission ended at : ${session.timerEnd}`)
```
RDIL, line 688:
```suggestion
                log(LogLevel.DEBUG, `Mission ended at: ${session.timerEnd}`)
```

components/eventHandler.ts
```diff
@@ -682,14 +677,23 @@ function saveEvents(
...
                 break
             case "ContractEnd":
                 if (!session.timerEnd) {
                     session.timerEnd = event.Timestamp
+                    log(
+                        LogLevel.DEBUG,
+                        `Mission ended at : ${session.timerEnd}`,
```
RDIL, line 695:
```suggestion
                        `Mission ended at: ${session.timerEnd}`,
```

components/profileHandler.ts
```diff
@@ -658,6 +671,71 @@ profileRouter.post(
...
+): Promise<void> {
+    const sessionId = save.ContractSessionId
+    const token = save.Value.LastEventToken
+    const slot = save.Value.Name
+    if (!contractSessions.has(sessionId)) {
+        throw Error("the session does not exist in the server's memory", {
+            cause: "non-existent",
+        })
+    }
+    if (slot in userData.Extensions.Saves) {
```
RDIL, line 696:
```suggestion
    const slot = save.Value.Name

    if (!contractSessions.has(sessionId)) {
        throw new Error("the session does not exist in the server's memory", {
            cause: "non-existent",
        })
    }

    if (slot in userData.Extensions.Saves) {
```

components/profileHandler.ts
```diff
@@ -658,6 +671,71 @@ profileRouter.post(
...
+        const delta = save.TimeStamp - userData.Extensions.Saves[slot].Timestamp
+
+        if (delta === 0) {
+            throw Error(
```
RDIL, line 700:
```suggestion
            throw new Error(
```

components/profileHandler.ts
```diff
@@ -658,6 +671,71 @@ profileRouter.post(
...
+                `the client is accessing /ProfileService/UpdateUserSaveFileTable with nothing updated.`,
+                { cause: "cause uninvestigated" },
+            )
+        } else if (delta < 0) {
+            throw Error(`there is a newer save in slot ${slot}`, {
```
RDIL, line 705:
```suggestion
            throw new Error(`there is a newer save in slot ${slot}`, {
```

components/types/types.ts
```diff
@@ -340,6 +366,13 @@ export interface UserProfile {
         }
         PeacockFavoriteContracts: string[]
         PeacockCompletedEscalations: string[]
+        Saves: {
```
RDIL, line 369: This can be null for existing profiles, null checks are needed wherever this is used.

RDIL (changes requested):

RDIL:
Mainly just consistency stuff, but one nullability concern.

In general, we use `new Error` instead of `Error`, and have a space before and after if statements, unless it starts/ends a block.

RDIL (approved):

RDIL (approved):
I assume you've tested this already, so LGTM

Issue #74, opened 2023-01-09
Item not included in peacock


cuocofra:
i have bought the trilogy pack on steam, but some items with peacock are not included, as far as i know is only one, the ica19 classicballer

The entP array in the user profile is
``` json
 ["1659040","1829592","1829593","1829594","1829600","1829601","1829602","1829603","1829604","1829605"]
```

moonysolari:
Original thread on Discord: https://discord.com/channels/826809653181808651/1060526886117199912

moonysolari:
Comments on `entP`:
1659040 is `STEAM_NAMESPACE_2021` (what is this?)
https://github.com/thepeacockproject/Peacock/blob/e548325955c00a2ed12546201352bae052f731f9/components/platformEntitlements.ts#L123
1829593 is `LEGACY_UNLOCKABLES`, which contains `FIREARMS_PISTOL_CLASSIC_SILVERBALLER`, but the check for this entitlement is currently broken. This might be the root cause of OP's issue
https://github.com/thepeacockproject/Peacock/blob/e548325955c00a2ed12546201352bae052f731f9/components/inventory.ts#L187-L195

All the other `entP`s are unknown and unchecked.

moonysolari:
@cuocofra  Have you installed any mods via simple mod framework?

cuocofra:
> @cuocofra Have you installed any mods via simple mod framework?

yes, but not on the first boot, the gun wasn't there even on the first boot

moonysolari:
You were right. This was indeed fixed by #75. Closing it now. @cuocofra please see if this bug persists on v6 when it releases, and if it does feel free to report.

PR #75, opened 2023-01-10
Fix a number of bugs on progression, entitlements, and warning logs

moonysolari:
### Progression fixes
- Clicking on the "Mission Stories" tab in-game now correctly shows a tick in the lower-right corner on the image of mission stories that have been completed before. 
- Adjusted the max level of a player from 5000 to 7500.
### Entitlement fix
- Now correctly gives ICA 19 Chrome if the player has the requiem pack. Previously it was the ICA 19 Classic Baller. See #74 
### Warning log fix
- Suppresses a useless "Contract undefined not found!" warning by adding a nullity check to the `resolveContract` function


RDIL (changes requested):
`allunlockables` is one of the few files that must remain consistent with the official version.

moonysolari:
> `allunlockables` is one of the few files that must remain consistent with the official version.

Where was the official version of it acquired?

moonysolari:
Reverted.

RDIL (approved):

PR #76, opened 2023-01-11
Fix a nullibility issue in opportunity data handling

moonysolari:
Fixed a nullibility issue introduced in #75 that causes contracts without opportunities to fail to start.

### DIFF
--- a/components/contracts/contractRouting.ts
+++ b/components/contracts/contractRouting.ts
@@ -329,12 +329,14 @@ function getContractOpportunityData(
         false,
     )
 
-    for (const ms of contract.Metadata.Opportunities) {
-        missionStories[ms].PreviouslyCompleted =
-            ms in userData.Extensions.opportunityprogression
-        const current = fastClone(missionStories[ms])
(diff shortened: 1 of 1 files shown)


components/contracts/contractRouting.ts
```diff
@@ -329,12 +329,14 @@ function getContractOpportunityData(
         false,
     )
 
-    for (const ms of contract.Metadata.Opportunities) {
-        missionStories[ms].PreviouslyCompleted =
-            ms in userData.Extensions.opportunityprogression
-        const current = fastClone(missionStories[ms])
-        delete current.Location
-        result.push(current)
+    if (contract.Metadata.Opportunities) {
```
AnthonyFuller, line 332: I think it would be better to have this as `if (!contract.Metadata.Opportunities) return result;` but that's just me.

AnthonyFuller (approved):
LGTM, just a comment that isn't really important.

RDIL (approved):
RSLGTM

PR #77, opened 2023-01-11
Update season 1 challenges and add challenges for the rest of the locations

moonysolari:
All challenge data were acquired with the `extract-challenge-data` script with no modifications.

**What works**
- Updated the season 1 challenges with some recently added ones.
- All destinations now have challenges shown in the Challenges tab on the location page.
- All contracts now have challenges shown in the Challenges tab on the planning page and in-game F1 page, including sniper ones.
- Players can complete challenges on sniper maps, and the completion data saves. (Tested with Psycho Daisies in The Last Yardbird)

**What might not work**
- I did not manually test every challenge and confirm that it works.
- Dartmoor Garden Show challenges are not working.

PR #79, opened 2023-01-12
Add Challenge completion and level completion percentage display

moonysolari:
- The "CHALLENGE COMPLETION" ratio and the "TOTAL COPMLETION" percentage now display the correct data at the bottom on the Destinations page, as well as on the page for each location. 
- The "number of challenges vs completed challenges" bar, "number of mission stories vs completed mission stories" bar, and completion percentage all display the correct data on the bottom right on the scoring page now.
- This PR is fully compatible with but does not depend on #77, including sniper missions.

### DIFF
--- a/components/scoreHandler.ts
+++ b/components/scoreHandler.ts
@@ -25,11 +25,12 @@ import {
     xpRequiredForLevel,
 } from "./utils"
 import { contractSessions, getCurrentState } from "./eventHandler"
-import { getConfig } from "./configSwizzleManager"
+import { getConfig, getVersionedConfig } from "./configSwizzleManager"
 import { _theLastYardbirdScpc, controller } from "./controller"
 import type {
     ContractSession,
     MissionManifestObjective,
+    PeacockLocationsData,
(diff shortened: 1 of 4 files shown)


components/scoreHandler.ts
```diff
@@ -261,9 +264,36 @@ export async function missionEnd(
...
+            locationChallenges,
+            userData.Id,
+            req.gameVersion,
+        )
+    const opportunities = contractData.Metadata.Opportunities
+    const opportunityCount = opportunities ? opportunities.length : 0
+    const opportunityCompleted = opportunities
+        ? opportunities.filter(function (ms) {
+              return ms in userData.Extensions.opportunityprogression
+          }).length
```
RDIL, line 295: Can probably be an arrow function to improve performance by a few cpu cycles 😆 
```suggestion
        ? opportunities.filter((ms) => (
              ms in userData.Extensions.opportunityprogression
          )).length
```

RDIL (approved):
Feel free to merge, suggestion is optional.

PR #84, opened 2023-01-19
Fixed issue where restarting Peacock would require to first go offline again

LennardF1989:
Recreated PR for further discussion after unintended merge into v6,


LennardF1989:
This is in v6 now.

PR #85, opened 2023-01-21
Fix completion data generation function

moonysolari:
The new `generateCompletionData` function in #30 returns the **parent location ID** in the `SubLocationId` field in its return value, which is not correct. It causes `'UI_LOCATION_PARENT_XXX_TEXT'` to appear in the mission end screen, among other places.  This PR restores the behavior that the **sublocation ID** should be in that field for most if not all purposes other than mastery. It also prevents location progression data from appearing in the bottom right for the sniper levels and the ICA facility, which is the correct behavior on official servers.

### DIFF
--- a/components/candle/masteryService.ts
+++ b/components/candle/masteryService.ts
@@ -112,7 +112,7 @@ export class MasteryService {
             XpLeft: nextLevelXp - locationData.Xp,
             Id: masteryData.Id,
             SubLocationId: subLocationId,
-            HideProgression: false,
+            HideProgression: masteryData.HideProgression || false,
             IsLocationProgression: true,
             Name: undefined,
         }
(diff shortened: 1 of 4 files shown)


RDIL (approved):

LennardF1989:
Looks good, but I feel instead of adding exceptions in code, we should just provide proper mastery data in the JSON files.

LennardF1989:
Alright, here's what I did:
- I added HideProgression as an optional boolean for the Mastery JSON
- I added a Mastery JSON for Facility to set the MaxLevel to 4 and HideProgression to true
- I changed the default CompletionData to be MaxLevel 1 and HideProgression true if a Mastery JSON is not found
- I respected the change you made to the masteryService by passing the subLocationId as we would have before the masteryService was a thing
- I added a fix to hide the Mastery-tab in the game if there are no drops associated with the Mastery JSON

The result is exactly as you hard-coded it, but with a lot more control and less code :)

Issue #16, opened 2022-11-13
Persistent shortcut challenges


RDIL:
- [x] Global challenges appear in missions they should not (did @AnthonyFuller fix this?)
- [ ] Make them actually completable
- [ ] Add dugong ones

AnthonyFuller:
>  Global challenges appear in missions they should not (did @AnthonyFuller fix this?)

Define this behaviour.

RDIL:
The ladders/door challenges for the Persistent Shortcuts are sent to the game even when not on the intended map, which is not the same behavior as the official servers.

AnthonyFuller:
Yes, this was fixed in the Evergreen branch by:
1. Following what inclusion data says (like we do with challenges, these systems should be merged no?)
2. Adding a location into their inclusion data, I assume what IOI do is check if they unlocked the actual challenge linked to it and send it that way, we can change that if wanted, (possibly adding our own inclusion data?)

RDIL:
Yeah.  Now all that remains is cherry-picking that behavior onto the main branch.

PR #88, opened 2023-01-21
Fix mission end screens for challenges

### LINKED ISSUE #12: Duplicate tiles of finished challenges at the end of the mission

moonysolari:
- Fixes #12 
- Challenge tiles that got "ticked" on a previous mission end screen will not be ticked in succeeding runs.
- Challenge tiles that did not get a chance to be "ticked" on a mission end screen will be ticked the next time a player triggers the mission end screen.
- Adds a `Ticked` field to every `ChallengeProgression` object in the player profile.
- Compatible with previous player profile saves.


components/profileHandler.ts
```diff
@@ -537,11 +537,13 @@ profileRouter.post(
         }
 
         for (const challenge of challenges) {
+            // TODO: Add actual support for shortcut challenges
```
AnthonyFuller, line 540: Care to elaborate?

AnthonyFuller (approved):
LGTM, just reply to my comment.

components/profileHandler.ts
RDIL: Pretty sure this is my to-do from a while back

moonysolari: Yeah it's #16

AnthonyFuller: Ah okay.

moonysolari:
Might be related to #86, as global challenges should also display similar behavior.

LennardF1989:
> Might be related to #86, as global challenges should also display similar behavior.
I reckon my thing at some point is going to combine with the missionEnd of scoreHandler as well (instead of acting as a plugin), so most likely, I will just have to grab "unticked" challenges to decide on the amount of XP to grant the user profile.

LennardF1989 (approved):

AnthonyFuller:
> I will just have to grab "unticked" challenges to decide on the amount of XP to grant the user profile.

Regarding challenges, XP is given AS SOON as they're unlocked. Like, pretty much instantly.

LennardF1989:
> Regarding challenges, XP is given AS SOON as they're unlocked. Like, pretty much instantly. No waiting for the end screen, during play.

That's fine, but I believe the original game will increment your Mastery XP progress bar in the first missionend as if you didn't have them before yet, no? If so, we also need something like an "original XP" before any challenges were applied. Unless we decide to just not do this in Peacock :P

AnthonyFuller:
> I believe the original game will increment your Mastery XP progress bar in the first missionend as if you didn't have them before yet, no?

It adds the XP onto mastery (and your player level) as soon as you complete the challenge, what you're thinking of is the visual part of it, which yes, occurs the first time after you've got that XP. So we'll likely need to have something like we have for ticked challenges which is a store of how much XP we need to show the player. This idea, of course, needs to be developed.

RDIL (approved):

Issue #90, opened 2023-01-26
Known issues in v6.0.0 alpha-1


moonysolari:
- [ ] H2 CAREER -> CHALLENGES page:
  -  [x] PATIENT ZERO: not working.
  -  [ ] Elusive challenges: not implemented.
  - [x] ICA SHIP: not showing.
-  [ ] Mastery:
   -  [ ] All games have the same mastery items as H3, which is incorrect for some locations. E.g. Hawkes Bay Lvl. 5 should not have Tanto in H2.
- [x] Some sublocation-specific challenges are not working on the planning page in H1 and H2. They do not correctly show the link to the mission under the challenge tile.
- [ ] #93

moonysolari:
Independent issues have been created for the items that need to be tracked in the long term. Closing this now.

PR #95, opened 2023-01-28
Fix a UI detail for the Challenges tab on the planning page

moonysolari:
- Removes the link to the corresponding mission under challenge tiles for the CHALLENGES tab on the planning page, to match the behavior on official servers. This bug was mentioned in #90.
- The link was previously not bugging in H3 but bugging in H2 (see screenshot). However, it should not be shown at all for the planning page.
- Whether the link is shown on other pages or not is not affected.
image


AnthonyFuller (approved):
LGTM.

PR #96, opened 2023-01-28
Fix the legacy contract downloader and improve options related to it

moonysolari:
- The `legacyContractDownloader` and `officialAuthentication` options toggle the same thing according to their descriptions. So, the latter is removed and the former is used in its stead.
- Fixed the legacy contract downloader for h3 by fixing the URL.
- Redid the logics regarding the contract downloaders such that 
  - H1 and H2 will always use the legacy contract downloader, regardless of the value of `legacyContractDownloader`, because this is the only downloader they can use.
  - H3 will use the legacy contract downloader if `legacyContractDownloader === true`, and the HITMAPS downloader otherwise. 
  - scpc will not use any downloader.
- Added a caveat for `userAuths` because it will only be non-empty if the legacy contract downloader is used.
- Removed redundant checks for location IDs because they no longer exist after #92.

### DIFF
--- a/components/controller.ts
+++ b/components/controller.ts
@@ -982,7 +982,11 @@ export class Controller {
         userId: string,
     ): Promise<MissionManifest | undefined> {
         const remoteService =
-            gameVersion === "h2" ? "pc2-service" : "pc-service"
+            gameVersion === "h3"
+                ? "hm3-service"
+                : gameVersion === "h2"
+                ? "pc2-service"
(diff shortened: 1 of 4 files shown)


AnthonyFuller (approved):
LGTM.

RDIL (changes requested):
I am against this change.  I can't remember the exact reason why, but it'll come back to me.

PR #97, opened 2023-01-29
Fix unwanted tiles on career -> challenges page 

moonysolari:
- Evergreen introduced a lot of sublocations with no challenges (e.g. `LOCATION_SNUG_PARIS_CRAPS`), which caused problems with the Career -> Challenges page.
- This PR improves the logic to leave these sublocations out when getting challenges for all locations.


AnthonyFuller (approved):
Whoops, didn't realise this would be an issue. Thanks!

PR #98, opened 2023-01-30
Implement contract history and completion tracker

moonysolari:
- A tick will now appear on the top-left of the contract tile if the contract has been completed before by the user. 
- A cross will appear instead if the contract has been attempted before by the player but never completed (i.e. failed).
- The "My History" menu now correctly displays the player's history of played contracts, sorted from most to least recent. 
- The "My History", "My Contracts", and "My Favorites" menus now support pressing E to only show completed, failed, or all contracts. All three views work. Previously configured views will persist between different user visits to the page.
- Tested to be working on all three game versions.


static/UserDefault.json
```diff
@@ -3,7 +3,7 @@
     "LinkedAccounts": {},
     "Extensions": {
         "PeacockEscalations": {},
-        "PeacockFavoriteContracts": [],
```
RDIL, line 6: Favorite contracts have their own system, this will break it.

RDIL (changes requested):

static/UserDefault.json
moonysolari: Could you please elaborate?

moonysolari: I added it back. Deleting it was an oversight.

RDIL (approved):

RDIL:
On standby until after v6 stable.

PR #99, opened 2023-01-30
Prevent disconnections caused by contract report requests

moonysolari:
- Previously, when hovering over a contract tile in the "My Contracts" section of the contracts menu, the user will disconnect if they press "F", which corresponds to reporting issues about the contract. 
- This PR prevents this disconnection by handling both requests related to this behavior. 
- It does not actually report anything to any server, but just returns an empty object to prevent disconnects.


components/index.ts
```diff
@@ -409,6 +409,10 @@ legacyRouter.use(
     "/authentication/api/userchannel/ContractSessionsService/",
     legacyContractRouter,
 )
+legacyRouter.use(
+    "/authentication/api/userchannel/ReportingService/",
+    legacyContractRouter,
+)
```
RDIL, line 415: Redundant: legacy router falls over to primary router if unhandled.

components/2016/legacyContractHandler.ts
```diff
@@ -105,6 +105,17 @@ legacyContractRouter.post(
...
+legacyContractRouter.post(
+    "/ReportContract",
+    jsonMiddleware(),
+    (
+        req: RequestWithJwt<never, { contractId: string; reason: number }>,
+        res,
+    ) => {
+        res.json({})
+    },
+)
```
RDIL, line 117: Redundant: legacy router falls over to primary router if unhandled.

RDIL (changes requested):

RDIL (approved):

RDIL (approved):

components/contracts/contractRouting.ts
```diff
@@ -324,6 +324,17 @@ contractRoutingRouter.post(
...
+contractRoutingRouter.post(
+    "/ReportContract",
+    jsonMiddleware(),
+    (
+        req: RequestWithJwt<never, { contractId: string; reason: number }>,
+        res,
+    ) => {
+        res.json({})
+    },
+)
```
RDIL, line 336: This will also attach /ReportContract to /ContractsService

RDIL (changes requested):

components/contracts/contractRouting.ts
moonysolari: Added a separate router for reports.

RDIL (approved):

components/contracts/contractRouting.ts
RDIL: Shouldn't this be removed then?

moonysolari: I forgor 💀 Removed now.

RDIL (approved):

Issue #106, opened 2023-02-02
Duplicate Ancestral Fountain Pens in the inventory


moonysolari:
Originally [reported](https://discord.com/channels/826809653181808651/1069589080578465872) by @NekoXIII.

`PROP_DEVICE_REMOTE_EXPLOSIVE_ANCESTRAL` has the same `RepositoryId` as `PROP_DEVICE_EXPLODING_FOUNTAIN_PEN_ANCESTRAL` and is believed to be causing this problem. Further investigation is needed to determine whether the former is a broken item or just an item with an incorrect `RepositoryId` in the files. I personally do not recall there being an "Ancestral Remote Explosive Device".

PR #112, opened 2023-02-03
Remove duplicated items in the user's inventory

moonysolari:
This PR fixes these duplicated items:
- The Ancestral Fountain Pen, mentioned in #106. 
- The Tactical Wet Suit in H2/H3. The cause is its two identical versions, one used as the default suit of Nightcall, and the other as the reward of The Mills Reverie. Only the latter appears in the inventory now.
- The Tactical Gear in H1/H2. It was loaded twice, once as `TOKEN_OUTFIT_COLORADO_HERO_COLORADOSUIT` and another time as `TOKEN_OUTFIT_COLORADO_HERO_RABIESSUIT`. Only the former appears in the inventory now.

### DIFF
--- a/components/inventory.ts
+++ b/components/inventory.ts
@@ -86,7 +86,15 @@ export function createInventory(
     // ts-expect-error It cannot be undefined.
     const filtered: InventoryItem[] = allunlockables
         .map((unlockable) => {
-            if (brokenItems.includes(unlockable.Guid)) {
+            if (
+                brokenItems.includes(unlockable.Guid) ||
+                // Remove TOKEN_OUTFIT_NEWZEALAND_HERO_NEWZEALANDSUIT
+                // Duplicate of TOKEN_OUTFIT_WET_SUIT, which is an unlock of The Mills Reverie challenge.
+                unlockable.Guid === "3a4bcbbf-d451-4853-8a85-649120e384df" ||
+                // Remove TOKEN_OUTFIT_COLORADO_HERO_RABIESSUIT
+                // Duplicate of TOKEN_OUTFIT_COLORADO_HERO_COLORADOSUIT.
+                unlockable.Guid === "e3234256-c061-48e6-b543-008d990affa1"
+            ) {
                 return undefined
             }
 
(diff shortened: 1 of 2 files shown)


components/inventory.ts
```diff
@@ -86,7 +86,15 @@ export function createInventory(
     // ts-expect-error It cannot be undefined.
     const filtered: InventoryItem[] = allunlockables
         .map((unlockable) => {
-            if (brokenItems.includes(unlockable.Guid)) {
+            if (
+                brokenItems.includes(unlockable.Guid) ||
```
RDIL, line 90: should just add to broken items

RDIL (changes requested):

components/inventory.ts
moonysolari: Added to brokenItems

components/ownership.ts
```diff
@@ -224,6 +224,14 @@ export const brokenItems = [
     "051825b9-44f2-4b3d-8ab1-8d4a07393c76",
     // eiffel tower melee weapon
     "a8a8f0da-a69b-428d-b8c1-faf8660ec318",
+    // PROP_DEVICE_REMOTE_EXPLOSIVE_ANCESTRAL (doesn't exist as of 3.140)
+    "099afc37-609b-48c9-9278-d3389b45829b",
+    // Remove TOKEN_OUTFIT_NEWZEALAND_HERO_NEWZEALANDSUIT
+    // Duplicate of TOKEN_OUTFIT_WET_SUIT, which is an unlock of The Mills Reverie challenge.
+    "3a4bcbbf-d451-4853-8a85-649120e384df",
```
RDIL, line 231: I'd prefer to remove the actual challenge unlock, as `TOKEN_OUTFIT_NEWZEALAND_HERO_NEWZEALANDSUIT` is the default suit for the location.  IOI doesn't have this problem because they don't show default location suits.

If we ever fix this on our end, this will need to be reverted.

moonysolari: Done.

RDIL (approved):

PR #117, opened 2023-02-04
Implement "Contract search", "Trending", and "Most played last week" tiles for contracts menu

### LINKED ISSUE #111: Support contract menus that require communication with official servers
Three contract menu tiles that require communication with official servers are currently unimplemented:
- [ ] Trending
- [ ] Most played last week
- [ ] Contract search

One solution is to send the request to the official servers to get a list of contracts to show to the user, but only download a contract itself when it is requested by a user.

moonysolari:
- Resolves #111.  The implementation uses the method mentioned in the issue. 
- Sends the `publicId`s of the contracts to the contract preservation backend whenever a list of contracts is acquired.
- `userAuth` is always saved as long as the game isn't scpc, since it is needed to get the contract lists from official.
- Tested to be working and have an acceptable request latency on all three games.


components/controller.ts
```diff
@@ -393,6 +398,10 @@ export class Controller {
      * Note: if you are adding a contract, please use {@link addMission}!
      */
     public contracts: Map<string, MissionManifest> = new Map()
+
+    // Converts a contract's repo ID to public ID.
+    public contractIdRepoToPublic: Map<string, string> = new Map()
```
RDIL, line 403: What do you mean a contracts repo ID? They don't have those.

moonysolari: I mean this. I know it's called "Id". I was trying to avoid confusion with "public Id". https://github.com/thepeacockproject/Peacock/blob/df5e3ae373f326317a89f071a98794c4d5732f56/contractdata/FEATURED/fortheseven-FC1.json#L299

RDIL: But this causes more confusion, as contracts are not part of the REPO.

RDIL: Additionally, this is just redundant, as we can do `resolveContract(id)?.Metadata.PublicId`.

moonysolari: Naming changed.

`resolveContract` doesn't work because it relies on having the `json` file of the contract locally, but here the public ID is needed to download the contract, so we don't have them locally yet.

components/menuData.ts
```diff
@@ -1547,7 +1555,76 @@ menuDataRouter.post(
...
+                "ContractSearchResponseTemplate",
+                req.gameVersion,
+                false,
+            ),
+            data: searchResult,
+        })
+    },
+)
+
+async function officialSearchContract(
```
RDIL, line 1580: This file is too big. We should move this elsewhere.

components/utils.ts
```diff
@@ -224,6 +233,8 @@ export function getDefaultSuitFor(location: string) {
 
 export const nilUuid = "00000000-0000-0000-0000-000000000000"
 
+export const hitmapUrl = "https://backend.rdil.rocks/partners/hitmaps/contract"
```
RDIL, line 236: `hitmaps` is spelled with an S

RDIL (changes requested):
A few more nits.

components/utils.ts
moonysolari: Fixed.

components/menuData.ts
moonysolari: To where? This function is only called by two other functions in this file...

RDIL: Somewhere else, I'm just tired of seeing this file get longer and longer.

moonysolari: Moved to `contractsModeRouting.ts`.

RDIL (approved):

PR #128, opened 2023-02-07
Get the latest entitlements for user upon authentication

### LINKED ISSUE #105: The Steam version of the game saves every 3 seconds on Peacock
[Reported](https://discord.com/channels/826809653181808651/1069457034589769778) multiple times. 

Possibly caused by `/authentication/api/userchannel/ProfileService/GetPlatformEntitlements`. If the game is not satisfied with the response to this request, it will send this request again and again. Each time it sends this request, it will also send one `/authentication/api/userchannel/ProfileService/SynchronizeOfflineUnlockables` and one `/authentication/api/userchannel/ProfileService/UpdateUserSaveFileTable` request. The latter causes the "Saving..." banner to pop in the bottom-right corner of the game. 

Exactly what the game is looking for in the response of `GetPlatformEntitlements` is unknown. I succeeded in reproducing this bug if I set the response to an empty array. More investigation is needed.

### LINKED ISSUE #119: New DLCs are not displayed without deleting `userdata`
When the user acquires new DLCs, Peacock will not know about it unless the user deletes `userdata`.

One solution is to get and overwrite the user's entitlements every time the user authenticates, and do not rely on past entitlement data.

moonysolari:
- Fixes #119.
- Possibly also fixes #105, which is related to outdated entitlements.
-  Added error try-catch blocks for Steam & Epic `getEntitlement` functions to detect failure scenarios.
- Tries to get the user's latest entitlements every time the user authenticates.
  - Upon success, overwrites the user's entitlement array.
  - Upon failure, 
    - For an old user, falls back to the previously stored entitlement array;
    - For a new user, stores an empty array as the entitlements.
- `castUserProfile` in `util.ts` no longer reports a missing entitlement array, since it is elsewhere handled.
- Tested on both Epic and Steam, for both new users and old users, with both succeeding and failing entitlement requests, and both up-to-date and outdated entitlement arrays. All yielded expected behaviors.
  - Failing entitlement requests were simulated by accessing a wrong URL, resulting in a 404 Not Found error.

### DIFF
--- a/components/entitlementStrategies.ts
+++ b/components/entitlementStrategies.ts
@@ -16,6 +16,7 @@
  *     along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 
+import { AxiosResponse } from "axios"
 import { log, LogLevel } from "./loggingInterop"
 import { userAuths } from "./officialServerAuth"
 import {
@@ -76,19 +77,28 @@ export class IOIStrategy extends EntitlementStrategy {
     override async get(userId: string) {
         // Note: Relies on the "legacyContractDownloader" flag.
         if (!userAuths.has(userId)) {
-            log(LogLevel.ERROR, `No data found for ${userId}.`)
+            log(LogLevel.ERROR, `No user data found for ${userId}.`)
             return []
         }
 
         const user = userAuths.get(userId)
 
-        const resp = await user?._useService<string[]>(
-            `https://${this._remoteService}.hitman.io/authentication/api/userchannel/ProfileService/GetPlatformEntitlements`,
-            false,
-            {
-                issuerId: this.issuerId,
-            },
-        )
+        let resp: AxiosResponse<string[]> = undefined
+
+        try {
+            resp = await user?._useService<string[]>(
+                `https://${this._remoteService}.hitman.io/authentication/api/userchannel/ProfileService/GetPlatformEntitlements`,
+                false,
+                {
+                    issuerId: this.issuerId,
+                },
+            )
+        } catch (AxiosError) {
+            log(
+                LogLevel.ERROR,
+                `Failed to get entitlements from Steam: got ${AxiosError.response.status} ${AxiosError.response.statusText}`,
+            )
+        }
 
         return resp?.data || []
     }
--- a/components/entitlementStrategies.ts
+++ b/components/entitlementStrategies.ts
@@ -16,7 +16,7 @@
  *     along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 
-import { AxiosResponse } from "axios"
+import { AxiosError, AxiosResponse } from "axios"
 import { log, LogLevel } from "./loggingInterop"
 import { userAuths } from "./officialServerAuth"
 import {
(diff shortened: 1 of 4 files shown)


components/entitlementStrategies.ts
```diff
@@ -76,19 +77,28 @@ export class IOIStrategy extends EntitlementStrategy {
...
+
+        try {
+            resp = await user?._useService<string[]>(
+                `https://${this._remoteService}.hitman.io/authentication/api/userchannel/ProfileService/GetPlatformEntitlements`,
+                false,
+                {
+                    issuerId: this.issuerId,
+                },
+            )
+        } catch (AxiosError) {
```
RDIL, line 96: In JS, this is a variable name and not a type. In addition, the error may not itself be an instance of AxiosError, it needs to be explicitly checked.

RDIL (changes requested):

components/entitlementStrategies.ts
ColossusChang: Added some checks on error type.

moonysolari: Added some checks on error types

lofcz:
Currently, this doesn't fix #105 (at least in my case). The schema of `GetPlatformEntitlements` response needs to be reversed in order to fix it.

moonysolari:
> Currently, this doesn't fix #105 (at least in my case). The schema of `GetPlatformEntitlements` response needs to be reversed in order to fix it.

Could you please elaborate on reversing the schema of the response of `GetPlatformEntitlements`? It isn't entirely clear to me what this means.

lofcz:
The game expects some fixed layout (schema) of `/authentication/api/userchannel/ProfileService/GetPlatformEntitlements` response. This branch (just like upstream v6) send the following body in this `POST` http request:

```js
{
    issuerId: this.issuerId,
}
```

And the game is not happy with it

AnthonyFuller:
The game sends that, nothing to do with Peacock.

components/platformEntitlements.ts
```diff
@@ -172,15 +173,37 @@ export async function getEpicEntitlements(
             .map((e) => `${namespace}:${e}`)
             .join(`&nsCatalogItemId=`)}`
 
-        return (await axios(url, v)).data as NamespaceEntitlementEpic[]
+        let result: NamespaceEntitlementEpic[] = undefined
```
RDIL, line 176:
```suggestion
        let result: NamespaceEntitlementEpic[] | undefined = undefined
```

components/platformEntitlements.ts
```diff
@@ -172,15 +173,37 @@ export async function getEpicEntitlements(
...
+                log(
+                    LogLevel.ERROR,
+                    `Failed to get entitlements from Epic: ${JSON.stringify(
+                        error,
+                    )}.`,
+                )
+            }
+        }
+
+        return result
```
RDIL, line 196:
```suggestion
        return result || []
```

RDIL (changes requested):
Strict null checks

components/platformEntitlements.ts
grappigegovert:
```suggestion
        let result: NamespaceEntitlementEpic[] = []
```

RDIL (approved):

janez33:
well I got the issue anyway. Installed latest fitgirl repack https://fitgirl-repacks.site/hitman-world-of-assassination/ and latest peacock 6.5.3
saving all the time
image
image

moonysolari:
No support for pirated versions including the fitgirl repack.

PR #147, opened 2023-02-27
Remove The Caged Bull from h1

### LINKED ISSUE #116: The Caged Bull is broken on H1
H1 does not have the resources that The Caged Bull requires. It should not be listed as a Featured Contract.

image

moonysolari:
Fixes #116.


RDIL (changes requested):
This is modifying the featured contract group in-place. Any modifications made to it will be global, as fcGroup will be a pointer to the group itself and not a copy of the group. It will need to be explicitly `fastClone`'d here.

RDIL (approved):

PR #148, opened 2023-02-28
Add mastery data for sniper missions

moonysolari:
## Main changes:
- Added mastery for all sniper firearms, including multiplayer ones. 
- Fixed sniper completion data to match with official. 
- Compatible with the [sniper co-op mod](https://www.nexusmods.com/hitman3/mods/444). 

## Other minor changes:
- Added the `plugins` folder to `.gitignore`.
- Removed the `Could not get CompletionData for location` debug message as it prints too many times for sniper locations.
- Fixed up some bugs in freelancer completion incidentally.

## Caveats:
- Changed the definition of a type and the function signatures for several functions.
- Progression on sniper mastery does not work yet. Players still get fully upgraded sniper rifles in-game.


components/candle/masteryService.ts
```diff
@@ -75,52 +75,78 @@ export class MasteryService {
...
-        const maxLevel = masteryData.MaxLevel || 20
+            userProfile.Extensions.progression.Locations[completionId]
 
         const nextLevel: number = Math.max(
             0,
             Math.min(locationData.Level + 1, maxLevel),
         )
-        const nextLevelXp: number = xpRequiredForLevel(nextLevel)
+
+        const nextLevelXp: number = completionId.includes("LOCATION")
```
LennardF1989, line 108: I think this is a sketchy bit of code. Using completionId for two different things, and then making an assumption like this. I much rather see a boolean flag or an enum to determine which of the two to use. On top of that, how does that work for the lines above when completionId is not a location? Aren't sniper levels also locations?

components/candle/challengeService.ts
```diff
@@ -371,12 +371,9 @@ export class ChallengeService extends ChallengeRegistry {
         const location = locations.children[child]
         assert.ok(location)
 
-        let contracts =
-            child === "LOCATION_AUSTRIA" ||
-            child === "LOCATION_SALTY_SEAGULL" ||
-            child === "LOCATION_CAGED_FALCON"
-                ? this.controller.missionsInLocations.sniper[child]
-                : this.controller.missionsInLocations[child]
+        let contracts = isSniperLocation(child)
```
LennardF1989, line 374: It's a shame the locations JSON doesn't have a field to designate it as a sniper level, this will prevent us from having to hardcode the location names.

components/menus/sniper.ts
```diff
@@ -109,19 +114,12 @@ export function createSniperLoadouts(
...
-                    XP: 0,
-                    Completion: 1,
-                    XpLeft: 0,
-                    Id: index,
-                    SubLocationId: "",
-                    HideProgression: false,
-                    IsLocationProgression: false,
-                    Name: index,
-                } as CompletionData,
+                CompletionData: controller.masteryService.getFirearmCompletion(
```
LennardF1989, line 120: I think the only problem with this is that it's not keeping in mind players that just want everything unlocked from the start?

components/candle/masteryService.ts
```diff
@@ -75,52 +75,78 @@ export class MasteryService {
...
         const userProfile = getUserData(userId, gameVersion)
 
-        //Gather all required data
-        const lowerCaseLocationParentId = locationParentId.toLowerCase()
-
-        userProfile.Extensions.progression.Locations[
-            lowerCaseLocationParentId
-        ] ??= {
+        // Generate default completion before trying to acquire it
+        userProfile.Extensions.progression.Locations[completionId] ??= {
```
LennardF1989, line 95: This is what I mean with the remark about the completedId being sketchy. A sniper level might use a progression key like `FIREARMS_SC_HERO_SNIPER_HM`, so won't this add non-existing locations to the player profile, for each sniper in the game? Also, what if two locations have the same sniper for its progressionKey, you will get progress of the wrong location, no?

LennardF1989 (commented):
Just some questions by inspecting the code through GitHub, I can do a more thorough one with it checked out later. Code-wise looks very clean, I just have a bunch of questions about some of the design decisions (mostly the progressionKey thing).

components/candle/masteryService.ts
moonysolari: I did this because I did not think adding an extra function argument is necessary because it can be inferred from the `completionId` passed. I can add it if needed.

When ` getLocationCompletion` is called on sniper locations, they will fail the "has mastery" check and return `undefined`, so `getCompletionData` won't be called on them. 

If the completionId is not a location, then it is a firearm. Each sniper mission has three firearms, one for each character, and they each have their own mastery, so that's nine masteries in total. The lines above will work for them as well because all it needs is an identifier for each mastery progression, which is the `completionId`. They will be stored as objects like `"FIREARMS_SC_HERO_SNIPER_KNIGHT":{"Xp":0,"Level":1}` in `Extensions.progression.Locations`. (Even though they are not locations. I know. But it would require a profile fix to change that name so I just went with it.)

components/candle/challengeService.ts
moonysolari: Yeah... but I guess it's the only way. Doubt they will add more sniper levels though.

components/menus/sniper.ts
moonysolari: For now everything **is** unlocked from the start. As for later when progression is added, I think it would be the same as opting out of normal progression. No matter which method you end up using, I will follow suit.

components/candle/masteryService.ts
moonysolari: > won't this add non-existing locations to the player profile, for each sniper in the game?
Exactly. For sniper missions, mastery is tied to the sniper rifles, and I just re-used the data structure that stores normal mastery progression. All nine snipers have unique `progressionKey`s and I think that was intended by IOI by calling it a "key".

components/utils.ts
```diff
@@ -118,6 +118,29 @@ export function xpRequiredForLevel(level: number): number {
     return level * 6000 - 6000
 }
 
+/**
+ * Get the number of xp needed to reach a level in niper missions.
```
RDIL, line 122:
```suggestion
 * Get the number of xp needed to reach a level in sniper missions.
```

components/utils.ts
```diff
@@ -118,6 +118,29 @@ export function xpRequiredForLevel(level: number): number {
...
+ */
+export function isSniperLocation(location: string): boolean {
+    return (
+        location.includes("AUSTRIA") ||
+        location.includes("SALTY") ||
+        location.includes("CAGED")
+    )
+}
+
 export function castUserProfile(profile: UserProfile): UserProfile {
```
RDIL, line 144: Oh boy, this thing. Hate it passionately, but that's for another day.

RDIL (changes requested):

components/candle/masteryService.ts
```diff
@@ -157,6 +155,9 @@ export class MasteryService {
                 gameVersion,
                 locationParentId.toLowerCase(),
                 masteryData.MaxLevel || DEFAULT_MASTERY_MAXLEVEL,
+                locationParentId.includes("SNUG")
```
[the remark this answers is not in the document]
moonysolari: Added an extra parameter for this.

RDIL (approved):

components/utils.ts
```diff
@@ -189,13 +189,35 @@ export const SNIPER_LEVEL_INFO: number[] = [
...
+ * @returns A boolean denoting the result.
+ */
+export function isSniperLocation(location: string): boolean {
+    return (
+        location.includes("AUSTRIA") ||
+        location.includes("SALTY") ||
+        location.includes("CAGED")
+    )
+}
+
```
AnthonyFuller, line 220: Question: `Contract["Metadata"]["Type"] == "sniper"`?

AnthonyFuller (changes requested):
One slight question.

components/utils.ts
moonysolari: But this function is for the location as a string? Do you mean I should consult `missionsInLocation` and then `resolveContract`? That's a long detour...

AnthonyFuller: Ah, I thought when you passed through the location it was from a contract. My apologies.

AnthonyFuller (approved):

PR #150, opened 2023-03-06
Fixed typo in condition for 'No Firearms' secondary task

Ferrster:
I think failing one task, but not the other, is not an intended behavior. Since Dartgun is not a **Fire**arm, it should not fail at all.


AnthonyFuller:
It's being kept as how it is on official servers. You can make a plugin to edit it on Peacock if you so wish.

Ferrster:
Understood, you're right, sorry. Completely forgot about possibility of plugins.

Issue #44, opened 2022-11-28
Option for log filters


NekoXIII:
The possibility to only have certain things logged in the start server.cmd window.
(I.E. Startup messages, mission end times)

PR #159, opened 2023-03-22
Add support for logging to console and files

LennardF1989:
Adds a logger based on [Winston](https://github.com/winstonjs/winston):
- Creates a new log every time Peacock is started (`frequency` is required, since the default value - `custom` - doesn't work as well).
- Keep a maximum amount of files (14 days right now, but can be tweaked)
- Files will have a JSON object per line with information: `{"timestamp":"02:50:58:732","level":"Info","message":"Booting Peacock internal services - this may take a moment."}`
- When debug mode is on, will also log `debug` and `trace` level, otherwise `info` and higer.
- Console still has colors and usual output

### DIFF
--- a/components/loggingInterop.ts
+++ b/components/loggingInterop.ts
@@ -22,6 +22,14 @@ import picocolors from "picocolors"
 import winston from "winston"
 import "winston-daily-rotate-file"
 
+const isDebug = ["*", "true", "peacock", "yes"].includes(
+    process.env.DEBUG || "false",
+)
+
+const isTest = ["*", "true", "peacock", "yes"].includes(
+    process.env.TEST || "false",
+)
+
 /**
  * Represents the different log levels.
  */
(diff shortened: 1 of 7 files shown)


components/loggingInterop.ts
```diff
@@ -19,6 +19,44 @@
...
 import picocolors from "picocolors"
+import winston from "winston"
+import "winston-daily-rotate-file"
+
+const isDebug = ["*", "true", "peacock", "yes"].includes(
+    process.env.DEBUG || "false",
+)
+
+const fileTransport = new winston.transports.DailyRotateFile({
+    filename: "logs/application-%DATE%.json",
```
LennardF1989, line 30: This should probably just be "peacock" instead of application.

components/loggingInterop.ts
```diff
@@ -19,6 +19,44 @@
...
+    datePattern: "YYYYMMDDHHmmss",
+    frequency: "1d",
+    maxFiles: "14d",
```
LennardF1989, line 33: Make this configurable, eg. by means of an environment variable.

LennardF1989 (commented):
Notes.

moonysolari:
When debug mode is off, can it be such that messages of all categories (including debug and trace level) will appear in files, while only those of info and higher will appear in the console?

LennardF1989:
> When debug mode is off, can it be such that messages of all categories (including debug and trace level) will appear in files, while only those of info and higher will appear in the console?

Yes, that's possible.

components/loggingInterop.ts
```diff
@@ -19,6 +19,44 @@
...
+    }),
+})
+
+const logger = winston.createLogger({
+    levels: {
+        error: 0,
+        warn: 1,
+        info: 2,
+        debug: 3,
+        trace: 4,
```
RDIL, line 55: Can we use the LogLevel enum here?

LennardF1989: Check the next commit - this is the best we can do with Winston.

LennardF1989:
> I want to do this in a way that works great and still looks pretty.

Well, got news, because the console output looks exactly the same as it was before!

LennardF1989:
This is now a "one logger to rule them all" solution. Preferable, you want a logger instance per service, which all tunnel down into the same place eventually.

The easiest would be adding an additional enum to the log with a list of category we can filter on, it's then relatively easy to turn them on/off as bitflags.

LennardF1989:
Latest commit has all you requested @RDIL and also integraties support for #44.

LennardF1989:
@moonysolari Yes, it does. Default log level for the file is TRACE and unless explicitly turned off, it logs everything even if the console doesn't.

RDIL (approved):

PR #161, opened 2023-03-23
feat: Initial support for mastery progression

J0k3r-1:
Initial support for Mastery Progression. It works for mastery unlockables, packages and deluxe items, but still need to figure out the other ones (which are still around 400 😥)

Going trough some of those unlockable acquisition methods on the wiki, most of them require x amount of contracts, elusive target completion, challenges (which I think it should be possible to track, so Ill look into it), and some don't even seem to be acquirable at all, like certain outfits, so need to think what to do with those as well


components/scoreHandler.ts
```diff
@@ -1019,6 +1020,15 @@ export async function missionEnd(
...
 
+    /**
+     * Clears inventory if new unlockables have been unlocked, so the inventory is recreated with them instead of fetching from the cache
+     *
+     * Im tempted to avoid doing this if the mission type is evergreen, because I dont think we need to clear inventory for freelancer, but would need to test (would save some cpu usage)
+     */
+    if (drops.length) {
+        clearInventoryFor(req.jwt.unique_name)
+    }
+
```
moonysolari, line 1031: Putting the performance issue aside, this way of doing unlocks will only grant unlockables at mission end screens. However, they should be granted to the user as soon as the user earns enough xp to unlock them, as per the behavior on official. Please take a look into the `Drop`s of challenges and mastery. 

A more ideal implementation might be:
- When booting, construct the inventory based on the user's entitlements, mastery progression, and challenge completion.
- Whenever a challenge is completed, check its `Drop` (if any) and grant it to the user.
- Whenever XP is gained, check the mastery level and grant corresponding `Drop`s to the user.

moonysolari (changes requested):
> x amount of contracts, elusive target completion, challenges
These are the unlockables locked behind challenges. Peacock currently do not have all the challenges that unlock items, but I've been adding them and eventually all will be added. 

> some don't even seem to be acquirable at all
If an item is not locked behind a challenge or mastery progression, or its challenge is not in Peacock yet, then it should be granted to the player by default. 

I have compiled a todo-list on unlockables and progression in #162, which you can use for reference. They need not (and should not) all come in one PR but please address the issue on when the unlockable is granted.

moonysolari:
Also please add a flag to opt out of progression.

J0k3r-1:
Thanks both, will look into it

J0k3r-1:
This should now handle the following:

[x] Support item unlocks for completed challenges.

[x] Add flag for toggling on/off item unlocks and progression.

[x] An unlock should be awarded to the player as soon as it is earned.

[x] Items that are not locked behind a challenge or mastery should be awarded to the player by default.

[x] Correctly handle loadout locks in Hokkaido and Miami (Loadouts are available for A Silver Tongue even before reaching mastery level 2 in Miami)

[x] Support mastery unlocks for ordinary locations, including stashes, starting locations and items.
 - Think most of it is handled, I've noticed some starting locations are for contracts and stuff, so they fall into the `otherItems` list, but they are appended to the inventory anyway, so unless some specific condition Im not aware of is required for them, it should be fine

components/flags.ts
```diff
@@ -105,6 +105,10 @@ const defaultFlags: Flags = {
         desc: "When set to true, all Freelancer unlocks will always be available.",
         default: false,
     },
+    enableMasteryProgression: {
+        desc: "When set to false, mastery progression will be disabled and all unlockabled will be awarded at the beggining",
```
RDIL, line 109:
```suggestion
        desc: "When set to false, mastery progression will be disabled and all unlockables will be awarded at the beginning",
```

components/menus/planning.ts
```diff
@@ -365,26 +365,27 @@ export async function planningView(
...
         const loadoutUnlockable = getUnlockableById(
             req.gameVersion,
-            parentLocationProperties?.NormalLoadoutUnlock,
+            parentLocationProperties.NormalLoadoutUnlock,
         )
 
         if (loadoutUnlockable) {
-            const locationMasteryData =
-                loadoutUnlockable &&
+            const loudoutMasteryData =
```
LennardF1989, line 375: ` loudoutMasteryData` should be `loadoutMasteryData` :)

components/inventory.ts
```diff
@@ -76,15 +99,95 @@ export function createInventory(
...
+        }, [])
+
+        const challengesUnlockables =
+            controller.challengeService.getChallengesUnlockables()
+
+        /**
+         * Seperates unlockable types and looksup for progression level
+         * on unlockables that are locked behind mastery progression level
+         */
+        const [unlockedItems, otherItems]: [Unlockable[], Unlockable[]] =
```
LennardF1989, line 130: I have a feeling this in combination with getMasteryForUnlockable can be done in a more performant way. /createInventory is called quite often when going through the menu, and constantly looping through ALL unlockable and looping through each mastery everytime is a big-O of n*n at worst (for each unlockable, loop through each mastery until we find what we need).

components/candle/challengeService.ts
```diff
@@ -1164,10 +1197,12 @@ export class ChallengeService extends ChallengeRegistry {
     }
 
     grantLocationMasteryXp(
+        gameVersion: GameVersion,
```
LennardF1989, line 1200: gameVersion is already in the contractSession.

LennardF1989 (changes requested):
Looks good so far :)

components/menus/planning.ts
J0k3r-1: I seriously need to get a spell check plugin lol

components/inventory.ts
J0k3r-1: Ive removed the clearInventory from the end mission, so this will only ever be once when the user logs in and fetched from cache every other time, so it should avoid any big-O operations. The drop awards are done by injecting the drops directly into the cache whenever its required

J0k3r-1: I do get the point for the getMasteryForUnlockable tho, I can probably map the data outside the functions, so its more easily accessable via objectIndex, so I dont have to loop trough everything again. Ill have a look

moonysolari:
I seem to directly get the freelancer weapon unlocks in my inventory (ancestral/ornamental weapons) while I am correctly locked out of the freelancer suits (golden contender, etc.)

J0k3r-1:
@LennardF1989 I misunderstood what you meant by using them together 😬 You're right, it would loop trough every mastery data on each unlockable, which adds too much complexity, even if the inventory is only created once. 
I've added another Map to the masteryService, as described in there, it does generate redundant data, but it makes accessibility to data much faster and with less complexion, so I think its a lesser evil.

J0k3r-1:
> I seem to directly get the freelancer weapon unlocks in my inventory (ancestral/ornamental weapons) while I am correctly locked out of the freelancer suits (golden contender, etc.)

Didn't realize there were actually freelancer unlockable for the inventory lol, will have a look tomorrow, should be an easy one to fix

J0k3r-1:
@moonysolari Seems like the evergreen gear unlockables for the inventory were actually being treated the same as packages, The Ids on the mastery data list are just the "package" for those items, so Ive added similar logic to the packages to actually handle awarding themcorrectly. Should be working now 👍

components/candle/challengeService.ts
```diff
@@ -1274,6 +1274,26 @@ export class ChallengeService extends ChallengeRegistry {
                     gameVersion,
                     unlockableIds,
                 )
+
+                // If missions type is evergreen, checks if any of the unlockables has unlockable gear, and award those too
+                if (contract.Metadata.Type === "evergreen") {
```
LennardF1989, line 1277: I'm pretty sure this is not how it works. Is this how it works. If you get to mastery level 5 of freelancer, the "EG_MASTERY_SKIN_LUXURIOUS_PISTOL" will be unlocked. This is not tied to any challenge in `_SNUG_CHALLENGES.json` - it just happens. So adding logic to handle this in challenges doesn't actually do anything. @moonysolari - thoughts? I feel we should add an optional "Drops" field to the Mastery Json type and check on that.

So:
```
{
  "Id": "EG_MASTERY_SKIN_LUXURIOUS_RIFLE",
  "Level": 10
}
```

Would become:
```
{
  "Id": "EG_MASTERY_SKIN_LUXURIOUS_RIFLE",
  "Level": 10,
  "Drops": [dropID]
}
```

AnthonyFuller: Everything in the mastery of Freelancer has an unlockable attached to it. All of them can be added to the inventory.

LennardF1989: There is also the "Gauze Suit" that I feel drops into your inventory after you die in Freelancer for the first time. So that's a whole other thing, again.

J0k3r-1: I think its just a consequence of this being on the challenges as we spoke before. As @AnthonyFuller mentioned, the gear unlocks from freelancer have an unlock id for the normal inventory, which is not really tied to the challenges, it just feels weird because where this is being done. We can move this to the progression service once that's done, suppose it will make more sense there.

Not sure about the suit tho, can investigate, but supposed its tied to a challenge that its not tracked yet, we would have to look into that first

AnthonyFuller: The Gauze Suit is a drop from a challenge.

moonysolari: > I feel we should add an optional "Drops" field to the Mastery Json type and check on that.

Sorry I'm a bit confused about why this needs to be added. I just tested and I can unlock The Ornamental Pistol on freelancer level 5. It appears in both the safehouse and in my inventory outside of freelancer, and I can take it to normal missions. Failing a mission also grants The Gauze Suit in my inventory. The items were also correctly locked before they were unlocked.

J0k3r-1: I think it may just be some confusion around the need for this. If it helps clarify the logic, effectively the ID for the freelancer version of the weapon is not the same as the one for the normal inventory, so unlocking only EG_MASTERY_SKIN_LUXURIOUS_RIFLE as per snug mastery list , will not unlock it on the inventory (also the reason why it was unlocked to begin with, because the inventory version was not behind mastery progression and so awarded to the user by default). 

Each evergreen gear type unlockable has an `Unlocks` field that references the ID for the inventory version, hence why the need to drill into the unlockables to unlock both version.

moonysolari: > Each evergreen gear type unlockable has an Unlocks field that references the ID for the inventory version, hence why the need to drill into the unlockables to unlock both version.

But I just tested and it seems both versions are already unlocked... Can you confirm?

moonysolari (approved):

AnthonyFuller (changes requested):
I do **strongly** think that if this gets implemented, it should be done "correctly" with a progression service instead of having something we will just have to re-write in the future.

I'm putting this as request changes to allow for a discussion to emerge.

J0k3r-1:
Added new progress service. It currently handles (most of it just basically moved out from the challenges service):

- [x] Awarding profile XP and level
- [x] Awarding mastery XP and level
- [x] Awarding mastery unlockables
- [x] Awarding challenges unlockables
- [x] Returns and creates (if necessary) mastery location data
- [ ] Hooks
  - As this is the first interaction, I left hooks outside of it (it should be fairly easy to implement later on), as we still need to decide which hooks we want to provide, and we need to consider what's allowed to modify validate those changes (e.g. We don't want the users to be able to manipulate drops, as it would screwup the XP requirements and potentially give the ability to give themselves deluxe items they don't own)

components/candle/progressionService.ts
```diff
@@ -0,0 +1,234 @@
...
+export class ProgressionService {
+    grantProfileProgression(
+        actionXp: number,
+        masteryXp: number,
+        challengeDrops: Unlockable[],
+        contractSession: ContractSession,
+        userProfile: UserProfile,
+    ) {
+        // Grants profile XP
+        this.grantUserXp(actionXp + masteryXp, contractSession, userProfile)
```
LennardF1989, line 48: Please introduce a `const xp` to make it very clear what XP means.

components/candle/challengeService.ts
```diff
@@ -1140,10 +1143,14 @@ export class ChallengeService extends ChallengeRegistry {
         //NOTE: Official will always grant XP to both Location Mastery and the Player Profile
         const actionXp = challenge.Xp || 0
         const masteryXp = challenge.Rewards?.MasteryXP || 0
```
LennardF1989, line 1145: Kind of in doubt here, but I would move any mention of XP and drops to the progressionService as well... It's kind of nice how `grantProfileProgression` can be called without a challenge.

Either way, please move the `//NOTE` as it's more applicable above the mentioned `const xp`.

components/inventory.ts
```diff
@@ -76,15 +100,130 @@ export function createInventory(
...
+            allunlockables
+                // Sorts packages and evergreen gear wrappers first, so related unlockables can be targeted after
+                .sort((_, b) =>
+                    b.Type === "package" ||
+                    (b.Type === "evergreenmastery" && b.Properties.Unlocks)
+                        ? 1
+                        : -1,
+                )
+                .reduce(
+                    (acc, unlockable) => {
```
LennardF1989, line 135: I feel we should move this to a local function for clarity, it's too nested now.

components/candle/progressionService.ts
```diff
@@ -0,0 +1,234 @@
...
+
+        // Grants Mastery Progression and Drops
+        this.grantLocationMasteryXpAndRewards(
```
LennardF1989, line 51: The old name was fine, we don't actually grant rewards in the function, it's all level and XP.

components/candle/progressionService.ts
```diff
@@ -0,0 +1,234 @@
...
+            masteryXp,
+            actionXp,
+            contractSession,
+            userProfile,
+        )
+
+        // Grants challenge related drops
+        awardDropsToUser(userProfile.Id, challengeDrops)
```
LennardF1989, line 59: `grantDrops` is better (it's literally in the comment). We're in the progressionService, it's pretty clear it's for the user.

LennardF1989 (changes requested):
Looking very good so far! Awesome work :)

Note: I only reviewed the masteryService and progressionService.

components/candle/progressionService.ts
J0k3r-1: Ive actually added the mastery drops award to the function, hence why I changed the name. I though about returning the drops and award them afterwards, but given they are tightly coupled with levelup from XP, I just kept them inside the function. I can move them out tho

J0k3r-1:
Added most of the suggestions, there's only one pending question👍

components/candle/challengeService.ts
```diff
@@ -1137,13 +1140,13 @@ export class ChallengeService extends ChallengeRegistry {
...
-        this.grantUserXp(xp, session, userData)
+        controller.progressionService.grantProfileProgression(
+            challenge.Xp,
+            challenge.Rewards?.MasteryXP,
+            challenge?.Drops ?? [],
+            session,
+            userData,
+        )
 
         writeUserData(userId, gameVersion)
```
LennardF1989, line 1151: This one can be removed since the grantProfileProgression will do that already.

components/candle/challengeService.ts
```diff
@@ -1137,13 +1140,13 @@ export class ChallengeService extends ChallengeRegistry {
...
-        //NOTE: Official will always grant XP to both Location Mastery and the Player Profile
-        const actionXp = challenge.Xp || 0
-        const masteryXp = challenge.Rewards?.MasteryXP || 0
-        const xp = actionXp + masteryXp
-
-        this.grantLocationMasteryXp(masteryXp, actionXp, session, userData)
-        this.grantUserXp(xp, session, userData)
+        controller.progressionService.grantProfileProgression(
+            challenge.Xp,
+            challenge.Rewards?.MasteryXP,
```
LennardF1989, line 1145: Please add back the fallbacks to 0, we want to pass on 0 in all cases, never an undefined.

components/candle/progressionService.ts
```diff
@@ -0,0 +1,238 @@
...
+    evergreenLevelForXp,
+    getMaxProfileLevel,
+} from "../utils"
+import { writeUserData } from "../databaseHandler"
+
+export class ProgressionService {
+    //NOTE: Official will always grant XP to both Location Mastery and the Player Profile
+    grantProfileProgression(
+        actionXp = 0,
+        masteryXp = 0,
```
LennardF1989, line 43: Set these to "number", not a default value. That makes it the responsibility of something else not to pass undefined values and expect things to work.

LennardF1989 (changes requested):
Last few (nitpick) things.

@moonysolari Can you give this a test-run? You are good at that :P

moonysolari:
I completed I'll Be Back and got the Gauze Suit, but I didn't get the Santa Suit after I completed 'Tis The Season... 

By the way, currently challenge pages do not show the unlockables associated with the challenges. Also the mission end page shows unlockables gained through mastery but does not show those gained through challenges. Do you think you can make these two work? It's also fine if you feel this does not fit in the scope of this PR.

J0k3r-1:
Fixed the challenge rewards not appearing on the missionEnd screen when unlocked

moonysolari:
> didn't get the Santa Suit after I completed 'Tis The Season

This is confirmed to be fixed. I can now get the Santa 47 Suit.

However there is a bug with unlocking firearms from freelancer masteries (e.g. the ornamental pistol). After I unlocked the ornamental pistol, I got this extra section in my inventory, while the pistol can also be found correctly under the pistols category: 
image
image

J0k3r-1:
That is...odd. Ill have a look

AnthonyFuller:
Either don't grant the main Evergreen unlock, or filter out `type === "evergreenmastery"`.

components/candle/progressionService.ts
```diff
@@ -0,0 +1,238 @@
...
+                return acc
+            }, [])
+            evergreenGearUnlockables.length &&
+                unlockables.push(
+                    ...getDataForUnlockables(
+                        gameVersion,
+                        evergreenGearUnlockables,
+                    ),
+                )
+        }
```
moonysolari, line 116: Yeah I think this part is what went wrong. Previously the `isEvergreenContract` passed was reversed, so this wasn't executed. Now with the `isEvergreenContract` fixed, this got executed and the unlockable got added to the inventory while it shouldn't. I think the game somehow manages to add firearms unlocked from freelancer mastery to both freelancer and normal game modes, so no need to add them again here.

moonysolari (changes requested):

components/candle/progressionService.ts
AnthonyFuller: The game adds them to the gear walls by checking what level you are.

Official does grant these to your inventory, but filters them from the inventory (for now at least). So that's what we should do!

J0k3r-1: Yeah sorry, I wasnt filtering those out on the createInventory so I assumed it was working, I didnt actually realized until now that they were actually being filtered out by the pre-existing filter function, so this was only a problem when awarding them because it injects directly on the cache. Should be addressed now 👍

AnthonyFuller: We should **still** grant them, but not show them, see the CTT branch for how this was done.

components/menus/planning.ts
```diff
@@ -286,6 +288,114 @@ export async function planningView(
...
+    ].filter(Boolean)
+
+    /**
+     * Handles loadout lock for Miami and Hokkaido
+     */
+    const parentLocationProperties =
+        locationData.parents[sublocation?.Properties?.ParentLocation]
+            ?.Properties
+
+    if (parentLocationProperties?.LimitedLoadout) {
```
moonysolari, line 372:
```suggestion
    if (sublocation?.Properties?.LimitedLoadout) {
```
Loadouts should be available for A Silver Tongue and Patient Zero even before unlocking loadout slots in Miami and Hokkaido, respectively. The properties of the sublocation should be used here.

J0k3r-1: ah!

components/candle/progressionService.ts
J0k3r-1: Ive added the awarding back, and filter on the function, same as createInventory. As I've mentioned tho, if the Official does have the evergreenmastery unlocks on the inventory, and we want to keep the same structure, we need to remove these filters from the inventory and filter them out at https://github.com/thepeacockproject/Peacock/blob/v6/components/menuData.ts#L360. This would keep them on the inventory as per official but not show them on the career inventory window. Let me know what you think

moonysolari (approved):

components/candle/challengeService.ts
```diff
@@ -141,6 +130,20 @@ export abstract class ChallengeRegistry {
...
     }
 
+    /**
+     * Returns a list of all challenges unlockables
+     */
+    getChallengesUnlockables() {
+        return [...this.challenges.values()].reduce((acc, challenge) => {
+            if (challenge?.Drops?.length) {
+                challenge?.Drops.forEach(
+                    (drop) => (acc[drop.Id] = challenge.Id),
```
moonysolari, line 140: This applies to everywhere that the `Drop` of a challenge is used.

With #174  merged, now the `Drop`s stored in challenge `json`s is a list of unlockable `Id`s instead. This change will affect the usage of the `Drop` array.

moonysolari (changes requested):

components/candle/challengeService.ts
J0k3r-1: Is should now take that into account

RDIL (approved):
Rubber-stamp admin approval for merge when everyone agrees it's ready.

moonysolari (approved):

components/menus/planning.ts
```diff
@@ -357,94 +467,7 @@ export async function planningView(
...
-                                      Properties: {},
-                                  },
-                                  type: i.Unlockable.Id,
-                                  owned: true,
-                              },
-                              IsContainer: true,
-                          },
-                      ].filter(Boolean),
+                contractData.Metadata.Type === "sniper" ? null : loadoutSlots,
             LimitedLoadoutUnlockLevel: 0, // Hokkaido
```
moonysolari, line 471: Consider editing the `LimitedLoadoutUnlockLevel` to be 20 for Hokkaido and 2 for Miami. It will be reflected in the UI.

components/inventory.ts
```diff
@@ -67,233 +91,376 @@ export function clearInventoryCache(): void {
...
-                    )
-                }
+/**
+ * Filters allowed unlockables
+ *
+ * @param gameVersion
+ * @param entP
+ * @returns boolean
+ */
+function filterAllowedContent(gameVersion: GameVersion, entP: string[]) {
```
LennardF1989, line 192: Not sure if we should do it as part of this PR, but this function just feels convoluted with all the hardcoded ID's and checks. Why not just get a datastructure in a file that allows us to do the same thing and keep the actual rules/configuration out of the code itself.

components/types/score.ts
```diff
@@ -31,6 +31,7 @@ export interface MissionEndChallenge {
 
 export interface MissionEndDrop {
     Unlockable: Unlockable
+    SourceChallenge?: Omit<MissionEndChallenge, "Drops">
```
LennardF1989, line 34: I rather just have a new interface for this.

LennardF1989 (changes requested):
Other than my comments, tested it yesterday combined with all other recent changes on main (please enable rebase merging) and works fine otherwise.

components/inventory.ts
AnthonyFuller: It's a part of Peacock already and prevents piracy. This code is pretty much exactly how it is on master, just in a different place. It's getting kept. There isn't really a good way to do it as:
1. Peacock **still** doesn't really properly support entitlements (and either way, they're not stored in the files).
2. We'd be adding data to existing unlockable entries. We'd rather not do that as it's then signficiantly harder to update.

The feature you suggest is out of scope for this PR in my opinion, and should be discussed and a new method be implemented separately.

components/menus/planning.ts
J0k3r-1: @moonysolari Is this to replace the filtering? Or to do together with it?

moonysolari: It has nothing to do with the filtering I think. On line 471 just set `LimitedLoadoutUnlockLevel` to 20 for Hokkaido, 2 for Miami, and 0 for everything else?

J0k3r-1: Yeah sorry, I though this was to lock the loadouts based on the level, I realize now its to display the required mastery on the loadout slot :) Should be addressed now 👍

LennardF1989:
Can you give this one final test @moonysolari?

moonysolari:
> Can you give this one final test

There was one problem which I have already fixed by pushing directly to the branch. Basically fixed Hokkaido not being able to start in H1, and also added Snow Festival `LimitedLoadoutUnlockLevel`.

Tested again and all seem to be in order.

LennardF1989 (approved):
Looks good to me, thank you very much for your efforts @J0k3r-1!

Just to confirm, because the game builds the inventory realtime, that would means that v6 profiles of players will get all the drops they should've got while they were playing without these fixes, right?

@RDIL Due to the large number of commits and some pretty elaborate changes from one to another, I guess we should do a squash and rebase for the cleanest result?

RDIL:
We can squash while merging.

J0k3r-1:
> Just to confirm, because the game builds the inventory realtime, that would means that v6 profiles of players will get all the drops they should've got while they were playing without these fixes, right?

Thats correct. The inventory is always recreated based on user progression and available entitlements, so it should work straight away with existing save files

AnthonyFuller (approved):
LGTM, thanks again for all your work towards this @J0k3r-1.

PR #164, opened 2023-03-26
Fix SA for freelancer

moonysolari:
SA wasn't properly fixed and I was still getting it.

### DIFF
--- a/components/scoreHandler.ts
+++ b/components/scoreHandler.ts
@@ -758,17 +758,6 @@ export async function missionEnd(
         timeTotal,
     )
 
-    let contractScore = {
-        Total: calculateScoreResult.scoreWithBonus,
-        AchievedMasteries: calculateScoreResult.achievedMasteries,
-        AwardedBonuses: calculateScoreResult.awardedBonuses,
-        TotalNoMultipliers: calculateScoreResult.score,
-        TimeUsedSecs: timeTotal,
-        StarCount: calculateScoreResult.stars,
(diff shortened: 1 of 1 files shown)


components/scoreHandler.ts
```diff
@@ -848,7 +848,7 @@ export async function missionEnd(
             }
         }
 
-        calculateScoreResult.silentAssassin = false
+        contractScore.SilentAssassin = false
```
LennardF1989, line 851: There better fix here is to set both to false, not replace one with the other. Also counts for the .star line below.

The same is true for the sniper contract start, it's also only modifying the calculateScoreResult, not both.

The better fix would be to move the `contractScore` to beyond the contract-type checks so we can keep modifying the calculateScoreResult. But we would need a way to signal the output to include it or not (eg. how we now set contractScore to undefined).

LennardF1989 (changes requested):

components/scoreHandler.ts
moonysolari: changed as requested

components/scoreHandler.ts
```diff
@@ -858,6 +847,17 @@ export async function missionEnd(
     let unlockableProgression = undefined
     let sniperChallengeScore = undefined
 
+    let contractScore = {
```
LennardF1989, line 850: I guess this could work, since the sniper game-mode doesn't use it. Please squash when merging so it's only this change :)

LennardF1989 (approved):

PR #168, opened 2023-03-29
Remove recon gear and update inventory items list

moonysolari:
- Removed Recon Gear, which is a dupe of the Tactical Turtleneck.
- Updated the lists `LEGACY_UNLOCKABLES` and `H2_RACCOON_STINGRAY_UNLOCKABLES`.


RDIL (changes requested):
Removing the recon gear breaks patient zero

moonysolari:
> Removing the recon gear breaks patient zero

Could you please elaborate? I just did a run of patient zero and there was nothing wrong with it.

RDIL:
In theory, it should break the planning screen for it, since the mandatory suit wouldn’t be in the inventory

moonysolari:
> In theory, it should break the planning screen for it, since the mandatory suit wouldn’t be in the inventory

It wasn't a problem and I could correctly start with that suit. Also in theory it shouldn't be a problem either. There are a lot of escalations with mandatory starting suits, but the suits are rewards of the escalations themselves and only appear in your inventory after playing them.

RDIL (approved):

PR #169, opened 2023-03-30
Update Elusive Target jsons to add VR bricks

moonysolari:
- For all ETs that have made an appearance in H3, their `json`s are updated with the latest official version. This includes `VR` bricks and a new `Codename_Hint` field.
- The `Id`s for some ETs have changed. This is not a result of mixing up Year 1 with Year 2 or 3, but rather a change in the `Id`s of the Year 1 versions themselves.
- For all ETs that have not made an appearance in H3, their `json` s are updated with the same `VR` bricks as the other ETs in the sub-location, and a new `Codename_Hint` field based on speculation.
- This is a side-product of my work on adding ET challenges. I do not have a VR device so I could not test to make sure all the missions now work in VR.


RDIL (approved):

RDIL:
Codename hints were always there, but we tend to strip them off because they are pointless (it's internal editor tooling).

moonysolari:
> Codename hints were always there, but we tend to strip them off because they are pointless (it's internal editor tooling).

Ah, I thought I'd add them because the main missions have them. What is this internal editor tool that you mentioned?

ahmed5029:
Thank you so much for doing this. Will this release in the next update or it is already updated. I have a VR device and can provide testing. I am a big fan of Hitman 3 vr and was looking for something like this. There is a Modder in nexus mod who added some levels where VR is not officially supported in hitman. Also, would you be able to have VR compatible with the new freelance mod of Hitman. Thanks for your efforts you and peacock you are amazing

moonysolari:
> Thank you so much for doing this. Will this release in the next update or it is already updated. I have a VR device and can provide testing. I am a big fan of Hitman 3 vr and was looking for something like this. There is a Modder in nexus mod who added some levels where VR is not officially supported in hitman. Also, would you be able to have VR compatible with the new freelance mod of Hitman. Thanks for your efforts you and peacock you are amazing

You're very welcome! This will come with the next release, and it's great to hear that you have a VR device and would be willing to playtest it once it's released. Freelancer VR is currently not scheduled. We might be able to add the VR bricks but I'm not sure if doing so will actually make VR work in Freelancer. Worth a try though.

Is [this](https://www.nexusmods.com/hitman3/mods/258) the nexus mod you mentioned? Does it work for you?

ahmed5029:
Yes, the mod I am referring to is called VRP extra assignments in VR and it works like a charm it is an offline mod though. I don't know if posted links here is allowed but here it is https://www.nexusmods.com/hitman3/mods/258. It works really well. Freelancing in VR would be a game changer to be honest. Let me know anytime you want VR tested for hitman and I can easily do so. Thanks for your hard work

PR #174, opened 2023-04-01
Fixed drops from not showing on their respective challenges in the menu

### LINKED ISSUE #171: Challenges not showing their drops in the menus
Only XP is shown.

LennardF1989:
Fixes #171.

### DIFF
--- a/components/candle/challengeService.ts
+++ b/components/candle/challengeService.ts
@@ -893,6 +893,23 @@ export class ChallengeService extends ChallengeRegistry {
         userId: string,
         isDestination = false,
     ): CompiledChallengeTreeData {
+        const allUnlockables = getVersionedConfig<Unlockable[]>(
+            "allunlockables",
+            gameVersion,
+            false,
+        )
+
+        const drops = challenge.Drops.map((e) =>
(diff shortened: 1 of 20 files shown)


LennardF1989:
Wow, this ESLint points out a weird one. The game actually works with this... Will look into what kind of magic is happening here.

AnthonyFuller (changes requested):
@RDIL wants to rejig what we store regarding drops, that probably wants to be addressed or considered before merging this. (I'm surprised this slipped through)

LennardF1989:
Wow, not sure how this happened. But when logging the supposed string array, it's actually a list of Unlockable objects. SavedChallenge is not the right type here.

RDIL:
yeah we were supposed to store strings but somewhere in the pipeline that just didn't happen 🤷

LennardF1989:
> yeah we were supposed to store strings but somewhere in the pipeline that just didn't happen 🤷

Fixed it proper now, testing. Will push in a moment.

RDIL (approved):

AnthonyFuller (approved):
LGTM

PR #179, opened 2023-04-04
Fix contract lookup for H2

### LINKED ISSUE #102: H2 bug when looking up an already downloaded contract
Previously got the following warning when trying to lookup an already downloaded contract on H2. Needs more investigation.
```
[Warn] Unable to save session [sessionID] because the session does not exist in the server's memory.
```

moonysolari:
Resolves #102.

Note that the two added json files are identical. It is possible to turn one of them into a symbolic link to avoid adding the file twice.


AnthonyFuller (approved):
LGTM, shame about having copies of the JSON, @RDIL will it get optimised out?

RDIL:
Nope

AnthonyFuller:
Okay then. I can't see any way around this unless we add some check either in the functions its used in or the `getVersionedConfig` function itself. Thoughts?

moonysolari:
Symbolic links?

RDIL:
It's not being in the repository twice, it's being in the bundle twice that is the problem.

grappigegovert:
You could do `import H2LookupContractTemplate from "../static/H2LookupContractTemplate.json"`
and then
```js
H2LookupContractByIdTemplate: H2LookupContractTemplate,
H2LookupContractFavoriteTemplate: H2LookupContractTemplate,
```
in the `configs` object definition.
I guess the template object still gets loaded into memory twice because of the foreach=>json.parse below that, but it's only in chunk0.js (and in the repo) once.

moonysolari:
> You could do `import H2LookupContractTemplate from "../static/H2LookupContractTemplate.json"` and then
> 
> ```js
> H2LookupContractByIdTemplate: H2LookupContractTemplate,
> H2LookupContractFavoriteTemplate: H2LookupContractTemplate,
> ```
> 
> in the `configs` object definition. I guess the template object still gets loaded into memory twice because of the foreach=>json.parse below that, but it's only in chunk0.js (and in the repo) once.

I think this is the best solution we've got and have implemented it.

PR #185, opened 2023-04-06
Improve export debug info

RDIL:
Improves the export debug info tool to include more data.

### DIFF
--- a/components/tools.ts
+++ b/components/tools.ts
@@ -32,7 +32,7 @@ import ProgressBar from "progress"
 import { resolve as pathResolve } from "path"
 import picocolors from "picocolors"
 import { Filename, PortablePath, ppath, xfs } from "@yarnpkg/fslib"
-import { ZipFS } from "@yarnpkg/libzip"
+import { makeEmptyArchive, ZipFS } from "@yarnpkg/libzip"
 
 // NOTE: make sure to update ALL 3 OF THESE VALUES, or things will break!!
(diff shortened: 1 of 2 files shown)


AnthonyFuller (approved):

components/tools.ts
```diff
@@ -112,10 +120,27 @@ async function exportDebugInfo(): Promise<void> {
...
+    await writeFile(zipFile, makeEmptyArchive())
+
+    const zip = new ZipFS(zipFile, { create: true })
+
+    await zip.writeFilePromise(zip.resolve("meta.json" as Filename), debugJson)
+
+    await copyIntoZip(zip, "logs")
+    await copyIntoZip(zip, "userdata")
+    await copyIntoZip(zip, "contractSessions")
+    await copyIntoZip(zip, "contracts")
```
moonysolari, line 137: I don't think `contracts` is really needed?

moonysolari (commented):
Maybe also contain all the plugins? Either the `*.js` files or just a list of the plugin names. Might need to look at both the `plugins` folder and under the main Peacock folder.

RDIL:
It contains the plugin names in meta.json

components/tools.ts
RDIL: In case of a contract specific problem? Not sure.

moonysolari (approved):

PR #187, opened 2023-04-06
Redo contract retrieval from official

### LINKED ISSUE #120: Add option to toggle Featured Contracts content
Add an option to toggle between displaying official Featured Contracts or Peacock exclusive ones.

moonysolari:
- Redid the contract retrieval logic. No longer uses the "publicId - contractId" lookup table to download contracts. Instead, contract definitions are fetched via `Planning` requests and are then loaded into the memory. This system has the following features, all of which are implemented in this PR:
  - Unclogs the `contracts` folder because it no longer stores the fetched contract jsons in there as files. This folder will again only store the contracts that the users put in there themselves. 
  - Automatically looks up a contract if its definition is not found, without need of its publicId.
  - ~~Supports fetching official featured contracts. When there are no more Peacock Featured Contracts to display on the Featured Contracts page, pressing "next page" will automatically fetch the latest official ones. This helps alleviate the problem that Peacock does not have enough featured contracts for users to complete all featured contracts challenges.~~
- Flags are resorted into categories to organize them better.
- Removed some outdated comments about authorization.


components/contracts/hitsCategoryService.ts
```diff
@@ -426,8 +443,14 @@ export class HitsCategoryService {
             const paginated = paginate(hitObjectList, this.hitsPerPage)
 
             hitsCategory.Data.Hits = paginated[pageNumber]
-            hitsCategory.Data.HasMore = paginated.length > pageNumber + 1
-            hitsCategory.CurrentSubType = `${category}_${filter}`
+            hitsCategory.Data.HasMore =
+                categoryName === "Featured"
+                    ? true
```
RDIL, line 448: What if we don't have more?

moonysolari: When we don't have more, the next page will display official featured contracts instead.

RDIL: What if we're out of official featured contracts?

moonysolari: When official ones are fetched instead, it won't reach this line. Check out the newly added `fetchFromOfficial` block above.

AnthonyFuller:
Some people use Peacock to get the json files of contracts from official, should they get stored in a sub folder instead of being kept in memory?

moonysolari:
> Some people use Peacock to get the json files of contracts from official

If you are talking about "contract lookup", then yes the contracts retrieved this way will still be downloaded into the `contracts` folder.

RDIL:
I’m going to take a closer look at this later, because I’m not 100% sure if this is doing what I think it’s doing.

RDIL (changes requested):
Blocking for now due to external factor

moonysolari:
Removed official featured contracts support.

RDIL (approved):

Issue #104, opened 2023-02-02
Game spams `/SaveEvents2` during Freelancer gameplay


moonysolari:
[Reported](https://discord.com/channels/826809653181808651/826809653744762932/1069405622073823272) a few times. The game spams `/SaveEvents2` during Freelancer gameplay. This is observed more often after exiting the level and before the mission end scene. This bug has happened before on v4 during normal gameplay but I don't know how it was resolved or by whom. I have not been able to reproduce this bug.
![image](https://user-images.githubusercontent.com/118079569/216217279-52cecf20-7f37-44dc-bd52-111de0027a5f.png)

LennardF1989:
I see this happen randomly at times. Perhaps we should introduce logic to make sure every event is always acknowledged, even if we don't use it. There is currently some logic in place to escape the event handler early on, and I'm pretty sure this causes this issue from time to time. Can't hurt to verify.

moonysolari:
This has not happened to me at all. I think we should add more debug output around events so that we at least know what event the game is sending when it happens.

moonysolari:
Closing this as completed by #190 and #191. Reopen if the problem reappears.

Kronzky:
It's still happening fairly regularly, and in all kinds on game modes (Freelance, Exclusive Targets, etc.).
I don't see anything specific triggering it, but the error is always `"The \"path\" argument must be of type string. Received undefined"`, after some time of playing (half an hour or so).
Sometimes it recovers, quite often it doesn't.

I've attached the logs of three crashes.
[crashes.zip](https://github.com/thepeacockproject/Peacock/files/11374752/crashes.zip)

AnthonyFuller:
Close the Peacock server. Then open `Tools.cmd`, press enter, then send the `DEBUG_PROFILE.zip` found in the Peacock directory.

moonysolari:
> I've attached the logs of three crashes.

I think your problem was caused by a cracked version of Peacock.

Kronzky:
> I think your problem was caused by a cracked version of Peacock.

The thing is, I'm trying to run this on a machine that has no Internet _at all_.
The stock Peacock is still connecting to the official servers at startup, so that won't work.

And I guess this error is caused by subsequent connection attempts during the game (which isn't an issue if you're connected anyway, but for a real offline machine this is causing problems).

I understand if you don't want to do support for a "cracked" versions, but to me it seems that still needing online access for running the game sort-of contradicts the purpose of this tool...

[...]

PR #190, opened 2023-04-06
Fixed an issue where refresh_token was not properly served

LennardF1989:
At some point the original JWT token will expire (even though we really don't care for the validity of the token) and the game desperately wants a new one. However, probably due to some refactors, this logic was in the wrong place.

Also added a test to make sure it can't break down again.


moonysolari:
Should/Does this fix  #104 ?

LennardF1989:
Yes, it should @moonysolari - most likely all related to that, this is all based on the few log files we received and the pattern has repeated a few times.

AnthonyFuller (approved):

PR #195, opened 2023-04-08
Added support for statistics under the Player Profile menu

LennardF1989:
Uses the same logic as the Career > Challenges, but without any additional filtering applied. Not sure if that logic was actually needed, but it also doesn't give unexpected results (I started in a different way, hence I felt like just adding that).

### DIFF
--- a/components/candle/challengeService.ts
+++ b/components/candle/challengeService.ts
@@ -17,6 +17,7 @@
  */
 
 import type {
+    ChallengeCompletion,
     ChallengeProgressionData,
     ChallengeTreeWaterfallState,
     ClientToServerEvent,
@@ -551,8 +552,6 @@ export class ChallengeService extends ChallengeRegistry {
             true,
         )
         const parent = locations.children[child].Properties.ParentLocation
-        const location = locations.children[child]
-        assert.ok(location)
 
         let contracts = isSniperLocation(child)
             ? this.controller.missionsInLocations.sniper[child]
@@ -1213,7 +1212,7 @@ export class ChallengeService extends ChallengeRegistry {
         challengeLists: GroupIndexedChallengeLists,
         userId: string,
         gameVersion: GameVersion,
-    ): { ChallengesCount: number; CompletedChallengesCount: number } {
+    ): ChallengeCompletion {
         const userData = getUserData(userId, gameVersion)
 
         userData.Extensions.ChallengeProgression ??= {}
@@ -1235,6 +1234,7 @@ export class ChallengeService extends ChallengeRegistry {
         return {
             ChallengesCount: challengesCount,
             CompletedChallengesCount: completedChallengesCount,
+            CompletionPercent: completedChallengesCount / challengesCount,
         }
     }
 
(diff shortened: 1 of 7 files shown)


components/candle/challengeService.ts
```diff
@@ -551,22 +553,30 @@ export class ChallengeService extends ChallengeRegistry {
...
-        }
 
-        return this.getGroupedChallengeLists(
-            {
+        let challengeFilterOptions: ChallengeFilterOptions
+
+        if (filterType === ChallengeFilterType.None) {
+            challengeFilterOptions = {
+                type: ChallengeFilterType.None,
+            }
```
moonysolari, line 562: Why is `None` needed here?

components/menuData.ts
```diff
@@ -1828,13 +1831,82 @@ menuDataRouter.post(
...
+    )
+
+    playerProfilePage.data.SubLocationData = []
+
+    for (const subLocationKey in locationData.children) {
+        //Ewww...
+        if (
+            subLocationKey === "LOCATION_ICA_FACILITY_ARRIVAL" ||
+            subLocationKey === "LOCATION_HOKKAIDO_SHIM_MAMUSHI" ||
+            subLocationKey.search("SNUG_") > 0
```
moonysolari, line 1853:
```suggestion
            subLocationKey.includes("SNUG_")
```

moonysolari (commented):
And how exactly does the behavior differ from official? Also have you fixed LOCATION_ICA_FACILITY_ARRIVAL reporting challenges

components/menuData.ts
LennardF1989: This was literally copied from your own logic, by the way, original I just had a startsWith :P

components/candle/challengeService.ts
LennardF1989: My initial frame of thought before I used the function way my own "getByLocation" function and I did not apply any other logic there other than locationId/parentLocationId check. Your filter also mentions it's only meant to be used for Career > Challenges. As such, not knowing exactly what it did, I felt it was best to not filter at all.

LennardF1989: https://github.com/thepeacockproject/Peacock/blob/673bfec718a1552e61474259fb16025b223386b8/components/candle/challengeHelpers.ts#L74-L75

components/menuData.ts
LennardF1989: https://github.com/thepeacockproject/Peacock/blob/5fff1c89ae739e063d713974a788af4bc38d82bb/components/menuData.ts#L240-L246

components/candle/challengeService.ts
moonysolari: I think you can give that filter a try. That comment was meant to record where it is currently used, not where it can only be used.

components/menuData.ts
moonysolari: Then I think both should use `includes`.

LennardF1989:
Reverted back to the original function, but added some additional logic to `getDestinationCompletion`, as the career page needs number of mission stories per SubLocation, unlike /Destination(s), which tallies them together (checked with Official to be sure that's intended behavior).

PR #196, opened 2023-04-09
Update AreaDiscovered challenges during evergreen missions

### LINKED ISSUE #193: Going to undiscovered map areas in Freelancer mode breaks that map's "Discover [location]" challenge
If you start a freelancer contract in a map where you have an undiscovered area, to which you then go to, you will "discover" it and the pin on the map will be replaced by the name of that area. However, this will not cause the "Discover [location]" challenge of the map to register that you have discovered that area, because this challenge is not present for Freelancer missions. This challenge will not be able to be completed, as a result. Currently the only way to resolve this is to do a full reset on all map area discoveries across all maps.

The game requests area discovery progress from the server with an `/authentication/api/userchannel/ProfileService/GetProfile` request, and the discovery progress is in `extensions.gamepersistentdata` in the response. It updates the discovery progress on the server with a  `/authentication/api/userchannel/GamePersistentDataService/SaveData` request. Neither request bears any information about the type of the contract. So we cannot try to cheat the game into thinking that the map is fully discovered only during Freelancer gameplay. 

Official has the same behavior. Currently the only solution I have for this problem is to ignore the `/SaveData` request, and update the area discovery progress when receiving an `AreaDiscovered` event via 
`SaveEvents2` instead. This is because the request of `SaveEvents2` comes with the contract session Id, which can be used to tell whether a session is an evergreen session. Area discovery progress is only saved when the session is not an evergreen session.

moonysolari:
Fixes #193 by ignoring `AreaDiscovered` events for evergreen missions. Areas discovered during freelancer gameplay will be reverted to undiscovered upon the next connection. Areas discovered normal contracts are still saved and counted towards the discovery challenges.

This behavior is different from official but it serves as a great alternative, and I think we should go with this solution unless someone wants to compile a list to look up which `PersistentBool` corresponds to which map for over 700 `PersistentBool`s


AnthonyFuller:
It's eaiser than you think, they're very well organised in the game's files.

moonysolari:
So Tony compiled the look-up table for which PersistentBool corresponds to which map. Now we need to decide where to update the challenge progression. One option is doing it upon getting an `AreaDiscovered` event, the other is doing it upon getting a `SaveData` request. The former is faster, while the latter is safer, as an `AreaDiscovered` event is a one-time thing while `SaveData` requests will be sent multiple times.

Personally I prefer doing it on `AreaDiscovered`.

moonysolari:
The following changes have been implemented to this PR:
- The previous approach has been scrapped.
- A map (`AreaMap.json`) has been added that maps an area ID to its corresponding discovery challenge ID.
- Upon an AreaDiscovered event, if the current session is freelancer, the corresponding discovery challenge will be informed of this event. 
  - This makes it possible to complete "Discover [location]" challenges (even solely) in freelancer.
  - The xp reward of that challenge will be awarded to the corresponding location, not to the safehouse. 
  - Previously the challenge was not informed because discovery challenges are not present in freelancer sessions. 
- `PersistentBool.json` has been removed, its purpose now carried out by `AreaMap.json`.

Tested and ready for review.

RDIL (approved):

AnthonyFuller (approved):
LGTM

PR #202, opened 2023-04-11
Support configurable default starting suits lock/unlock

### LINKED ISSUE #200: Add a flag to toggle whether default starting suits are automatically awarded to the player
We should add a flag to toggle whether normal starting suits are added to the inventory and it should be defaulted to false, because:
- Too many "Subject 47 suit not working" questions received.
- Some of them are rewards (Tactical Turtleneck, wetsuit), awarding them directly defeats the purpose.
- They clog up the inventory, especially when the gloved/hatted variants are also present.

Note that when they are not in the inventory, it must still be possible to choose them in their respective locations.

moonysolari:
- Resolves #200.
- Added a flag to toggle whether default starting suits are awarded to the player.
- When default starting suits are awarded AND mastery is enabled, the suits that are challenge/mastery rewards will still be locked.
- When default starting suits are not in the inventory, it is still possible to choose them in their respective sub-locations.
- The Author and The Icon now correctly default their starting suits to "Italian Suit Without Sunglasses". Previously it was the version **with** sunglasses.
- Works for all game versions.


moonysolari:
I would prefer this be merged before the next release, because on the current version the default suit for Ambrose Island is missing, due to the recent addition of mastery progression locking up The Guerilla Wetsuit.

image

components/inventory.ts
```diff
@@ -384,13 +385,72 @@ function filterAllowedContent(gameVersion: GameVersion, entP: string[]) {
...
+
+    // We need to add a suit, so need to copy the cache to prevent modifying it.
+    const newInv = [...inv]
+
+    // Yes this is slow. We should organize the unlockables into a { [Id: string]: Unlockable } map.
+    const locationSuit = getVersionedConfig<Unlockable[]>(
+        "allunlockables",
+        gameVersion,
+        true,
+    ).find((u) => u.Id === getDefaultSuitFor(sublocation))
```
AnthonyFuller, line 418: @RDIL thoughts on doing this?

RDIL: Unless we refactor the whole codebase into using a map for this, the time spent generating a map for this will probably be the same tbh

AnthonyFuller: Yeah, I just edited my original message adding that we'd need to change anything that uses it to use `Object.values`. We could also just change the static file to a map.

AnthonyFuller: I've actually just had a look at everywhere that uses it, and it would probably reduce the number of filter/find calls we have to do, **but** if an SMF unlockables is used, we'd have to generate a map for it.

AnthonyFuller (approved):

PR #209, opened 2023-04-14
Add support for Elusive Target Arcade

AnthonyFuller:
This PR adds support for Elusive Target Arcade.

What's left to do/fix (feel free to add more):
- [x] Incorrect challenges (elaborated below)
- [x] Arcade challenges in general
- [x] Failing at one of the later ETs in the series does not reset you back to the first ET.
- [x] Global arcade challenges added, but not working yet.
- [x] Arcade challenges don't show in the Career - Challenges page.
- [x] Arcade challenges are not correctly filtered in the Destination - Challenges tab. (Correct behavior: if the location has an ET, then all global arcade challenges are shown. Additionally, show challenges of arcade contracts assigned to this location.)

**Incorrect challenges**
Currently, we use the group contract's location ID for challenges/mastery, which is fine for the main gamemode page, but not for the contracts as they span different locations so this means the planning page has incorrect challenges for the contract location, as well as in-game challenges (i.e. the ones being tracked), and the final mission end challenges/mastery.

IOI also at some point changed from setting the group location to the first level's location to the final level's location (which being honest, makes more sense).


moonysolari:
Issue: failing at one of the later ETs in the series does not reset you back to the first ET.

AnthonyFuller:
> Issue: failing at one of the later ETs in the series does not reset you back to the first ET.

We're not going to support the 12 hour lockout, but I'm not sure if we should reset it either on fail. I've put it to a poll and it will conclude in 24-48 hours.

RDIL (approved):

RDIL (approved):

components/controller.ts
```diff
@@ -593,7 +593,7 @@ export class Controller {
     }
 
     private getGroupContract(json: MissionManifest) {
-        if (json.Metadata.Type === "escalation") {
+        if (["escalation", "arcade"].includes(json.Metadata.Type)) {
```
LennardF1989, line 596: This array comes back a few times and can therefore be considered magic values. Can we move that to a constant somewhere?

components/contracts/elusiveTargetArcades.ts
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * These are the parent locations with sublocations that have ETAs,
+ * and thus need to show ETA challenges on their challenges page.
+ *
+ * It is possible to determine whether a parent is such a parent location
+ * when filtering out challenges for the parent location, but it is
+ * much more efficient to do it this way.
+ */
+export const parentsWithETA = [
```
LennardF1989, line 82: My fingers are itching by this bit and `locationsWithETA`, seems like premature optimization.

LennardF1989 (commented):
Looking good!

components/controller.ts
AnthonyFuller: Sure, this has been done.

components/contracts/elusiveTargetArcades.ts
AnthonyFuller: I'll defer this to @moonysolari.

moonysolari: Both of these arrays are used solely in the `filterChallenge` function. When getting the challenges for a contract or a location, this function is called on every challenge that is either in the parent location or a global challenge.

To do without `locationsWithETA`, we will need to loop over all arcade levels and run `resolveContract` to get their locations every time this array is needed.

To do without `parentsWithETA`, we will need to loop over all locations in `locationsWithETA` and get their parent locations from the config file every time this array is needed.

I wrote them as constants considering the frequency of `filterChallenge` being called, the complexity of acquiring the locations with ETA in real-time, and the likelihood that this list will change. However, this implementation is still up for discussion.

LennardF1989: But why not build it once at runtime, instead of hardcoding it? This is going to give issues if someone is going to add custom stuff through plugins.

moonysolari: Good idea. Addressed as required.

AnthonyFuller: You also need to remove the arrays from `elusiveTargetArcades.ts`.

LennardF1989: LGTM - do we want to get rid of the whole orderedET array as well and just built that runtime too?

AnthonyFuller: We can't, there's no ordering data for them.

LennardF1989: In that case I sign off on this in its current state, good job all :)

PR #219, opened 2023-04-16
Add support for all imports to plugins

LennardF1989:
After #218 I made a plugin for someone on Discord and learned the hard way that compiling from TS to JS is not as easy as it seemed...

The underlying issue is that the @peacockproject/core package contains all definitions (if you would add it as a dependency), but only a **very limited** subset was actually allowed to be used.

This PR fixes that issue by adding an additional step to the build pipeline to generate a full "require table" that can be used by plugins, while still allowing to add custom things to the require table as was possible. It should be backwards compatible with existing plugins.

By doing this, I also found an issue where an export in profileHandler was using something that was not exported in itself, so this new logic can also act as a sanity check. Last but not least, I added `PEACOCK_DEV` as a requirement before Peacock would try to load TypeScript-based plugins, otherwise running as a non-development version would cause (harmless, but annoying) load-errors.

Side-note: The reason I added it as an extra if-check to the "require"-method itself is to prevent an issue where the variable containing all generated imports is not defined yet (because the file depends on pretty much all other files, it will be near the end of chunk0.js, so way after the `peacockRequireTable` already initialized).

### DIFF
--- a/components/controller.ts
+++ b/components/controller.ts
@@ -20,7 +20,6 @@ import { existsSync, readdirSync, readFileSync } from "fs"
 import { readdir, readFile, writeFile } from "fs/promises"
 import * as atomically from "atomically"
 import { join } from "path"
-import * as dataGen from "./contracts/dataGen"
 import {
     generateUserCentric,
     getSubLocationFromContract,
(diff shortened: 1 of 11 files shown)


RDIL (approved):
Great changes!

PR #223, opened 2023-04-17
New options.ini format

LennardF1989:
- Added support for sections to flags
- Added support for plugic-specific flags

Currently in draft, the `Flag` type will get some more extensions for the menu integration I'm working on, also contemplating just updating `getFlag` to have two arguments. Or for backwards compatibility add a plugin-specific `getFlag` to force plugins to use the right thing from the start and slowly get rid of `getFlag` with just 1 arg.


LennardF1989:
Retiring this in favor of the in-game options menu, as that heavily relies on this.

PR #225, opened 2023-04-17
Add support for Career -> Mastery page

AnthonyFuller:
As title.

EDIT: I should note that this only really works for H2/3, H2016 does not have `/profiles/page//MasteryLocation?` handled, plus H2016 has pro mastery which I don't believe we have implemented. For H2016, the tiles don't properly display mastery progress.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,5 @@ plugins
 logs
 
 THIRDPARTYNOTICES.txt
+
+overrides
--- a/components/tools.ts
+++ b/components/tools.ts
@@ -34,14 +34,18 @@ import picocolors from "picocolors"
 import { Filename, PortablePath, ppath, xfs } from "@yarnpkg/fslib"
 import { makeEmptyArchive, ZipFS } from "@yarnpkg/libzip"
 
-// NOTE: make sure to update ALL 3 OF THESE VALUES, or things will break!!
-const IMAGE_PACK_BIN =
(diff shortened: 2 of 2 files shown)


AnthonyFuller:
Just added support for H2016's Career -> Mastery page, making my edit in my original post invalid. Apart from the fact that as game-specific mastery isn't implemented (alongside pro1 specific mastery), that right now the levels are invalid.

components/candle/masteryService.ts
```diff
@@ -80,10 +83,9 @@ export class MasteryService {
         gameVersion: GameVersion,
         userId: string,
     ): MasteryDataTemplate {
-        const subLocation: Unlockable = getSubLocationByName(
-            locationId,
-            gameVersion,
-        )
+        const location: Unlockable =
+            getSubLocationByName(locationId, gameVersion) ??
+            getParentLocationByName(locationId, gameVersion)
```
LennardF1989, line 88: I have mixed feelings about this code. I feel we have multiple ways of achieving the same result now. Do you actually need the `template` that is part of this response, or are you only interested in the `MasteryData` - in that case, you should just call getMasteryData directly.

Can we check if all usages of these two functions is consistent now over the codebase of Peacock?

AnthonyFuller: This function is only used by `MasteryDataForLocation`, the Career -> Mastery page requests this endpoint _by parent location_ thus is the reason for adding parent location handling to this function.

LennardF1989:
LGTM

PR #233, opened 2023-04-20
Update locale_h3.json

Wuzmart:
Russian localization of H3 strings


AnthonyFuller (changes requested):
Thanks for these! But could you also copy them over to `locale_h2.json`. Should just be a simply copy paste.

AnthonyFuller (approved):

PR #235, opened 2023-04-20
Tick all unticked challenge in parent location on the mission end page

### LINKED ISSUE #231: Mastery progression rewards are not displayed correctly on the mission end screen
Currently the following may happen:

Using a fresh profile, go to Club 47 and complete a lot of challenges such that the Bangkok mastery progresses, but do not trigger the mission end screen. Now go to The Source and trigger the mission end screen. The rewards corresponding to the mastery levels that progressed as a result of the challenges completed in Club 47 will not be shown. 

This is because currently the `oldLocationXp` on the mission end page is not organized correctly. It is computed using the current XP and the XP gain. The XP gain is calculated using the unticked challenges tracked in this session. However, challenges not in the scope of the current contract may have contributed to the progression. 

One solution is to keep a `PreviouslySeenXp` for each parent location, and update upon mission end pages. This is also reflected in the default user profile. 

https://github.com/thepeacockproject/Peacock/blob/7169d624a87d2ab351ef83f8719ace5110ce67b5/static/UserDefault.json#L17-L24

moonysolari:
On the mission end page, all challenges in the parent location should be ticked, according to the behavior on official.

Resolves #231.


moonysolari:
This PR is now ready for review. Apart from the comment above, this PR also
- Made location mastery aware of global challenges, such that when only global challenges are completed in a run, and the XP gained through them caused mastery to progress, mastery progression rewards will be correctly displayed on the end screen.
- Fixed an issue where the freelancer progression bar would display negative values
- Updated the default profile.

static/UserDefault.json
```diff
@@ -15,14 +15,14 @@
             "LastScore": 0,
             "LastCompletedChallenge": "00000000-0000-0000-0000-000000000000",
             "Locations": {
-                "location_ica_facility": {
+                "location_parent_ica_facility": {
```
AnthonyFuller, line 18: Applies to all key changes in this file.

I don't see existing userprofiles being converted from sublocation -> parents, this **must** be done.

AnthonyFuller (changes requested):
LGTM so far... I see a glaring issue though.

static/UserDefault.json
moonysolari: Existing user profiles already use parents. They don't use sublocation at all. That's why I made that change.

https://github.com/thepeacockproject/Peacock/blob/b3d0353df1b69740e9c7c9b2417bd686ee1e36b8/components/candle/masteryService.ts#L186

AnthonyFuller: Ah I see.

AnthonyFuller (approved):

PR #246, opened 2023-04-22
Update locale.json

Wuzmart:
Added even more Russian strings


AnthonyFuller (approved):
Thanks for the translations!

PR #254, opened 2023-04-26
Fix ET challenges not ticked on the mission end page

moonysolari:


moonysolari:
Also fixed elusive targets SA challenges not working. Ready for review.

AnthonyFuller (approved):

PR #269, opened 2023-05-06
Updated the Khakiasp Documentation

riisikumi:
Updated the Maelstrom photo objective on levels 2 and 3 of the Khakiasp Documentation to let the player choose which photo to pick up, since it isn't very clear that the current objective requires a specific photo


resources/locale.json
AnthonyFuller, line 1: Please add the new English version to **all** languages, overriding any previously done translations.

AnthonyFuller (changes requested):

AnthonyFuller (approved):
Thanks!

PR #272, opened 2023-05-08
Fix disconnection when dying in arcade

moonysolari:
Fixed disconnecting when dying in arcade because the event will have a non-string Value.
``` json
"Value": {
    "FailType": "HitmanDead",
    "Reasons": ["Agent Down"]
}
```


AnthonyFuller (approved):
Whoops.

PR #280, opened 2023-05-18
Add new Peacock exclusive challenges for Season 2 ETs

moonysolari:
Added a new Peacock exclusive challenges for each Season 2 ET. Completing all such challenges unlocks the Solstice Suit.


AnthonyFuller:
I'm still a bit unsure about what IOI are going to do about the Solstice suit, hence why I'm hesitant to merge this. Please fix the merge conflict while I think more about it.

PR #281, opened 2023-05-21
Allow the patcher project to open and build on Visual Studio 2022

prolium:
.NET Framework 4.0 is deprecated in vs2022 and the targeting packs are no longer available to download from Microsoft.
Targeting v4.6 allows it to be opened and built, also v4.6 was an in-place update for v4.0-v4.5.2: [Microsoft .NET Framework 4.6 (Web Installer)](https://www.microsoft.com/en-gb/download/details.aspx?id=48130)


LennardF1989:
.NET 4 technically allows compatibility with Windows XP. The next best choice would be 4.5, since that ships with Windows 8 by default. Only starting Windows 10 is .NET 4.6 installed by default.

Not saying this is wrong, but something to consider.

AnthonyFuller:
> .NET 4 technically allows compatibility with Windows XP.

The game can't even run on XP (and the server itself can't run on Win7), so this is definitely not a concern if it ever even was one.

RDIL:
Dropping XP support isn't a concern at all. My only concern would be if the relevant runtimes are shipped by default on Win10 and Win11. If not, and this is another download, then I'm -1 on this.

RDIL:
Looks like the CI doesn't ship with the TP by default either, I don't have a ton of time to dive into the image and see it's contents though so I'm not sure how to proceed.

RDIL:
Thank you for the PR btw! This is something I've been meaning to do for a few months now 😄

TheBoxyBear:
Where can I acquire the targeting pack? It also works in 4.6 for me but I would have to downgrade before making a PR until this is merged.

RDIL:
@TheBoxyBear the 4.0 targeting pack is shipped with the VS2019 installer - https://learn.microsoft.com/en-us/visualstudio/releases/2019/redistribution#vs2019-download

RDIL (approved):

grappigegovert:
I trust that the CI works as intended; looks fine otherwise 👍

grappigegovert (approved):

AnthonyFuller (approved):
I concur with @grappigegovert.

RDIL:
Awesome, thank you @prolium!

PR #291, opened 2023-06-14
[perf] Inventory rework (part 1)

RDIL:
Changes the inventory functions to use a cached map.


AnthonyFuller (approved):
Fucking finally.

RDIL:
Some benchmarking we need to do (as performance issues are a problem in this case):

- Currently, createInventory stores the created inventory as a list of inventory objects. This will use a lot of memory. What is the overhead of storing that as a string[] instead, and then translating to actual objects when needed? Which is faster?

components/utils.ts
```diff
@@ -520,50 +520,7 @@ export function addDashesToPublicId(publicId: string): string {
...
-                    result[i] = fastClone(item[i])
-                }
-            }
-        } else {
-            result = item
-        }
-    }
-
-    return result
+    return structuredClone(item)
```
RDIL, line 523: New built-in added in Node 17

AnthonyFuller:
> Some benchmarking we need to do (as performance issues are a problem in this case):
> 
>     * Currently, createInventory stores the created inventory as a list of inventory objects. This will use a lot of memory. What is the overhead of storing that as a string[] instead, and then translating to actual objects when needed? Which is faster?

Well isn't it being stored _twice_ due to the config `allunlockables`? So if we can just "unload" that when it's been placed into a map, then we can just eat up the memory usage of the map. We knew the map would be an issue anyway.

RDIL:
@AnthonyFuller there isn't a really good way of unloading it. No matter what we really do, it will still be in memory because of sourcemaps and such

AnthonyFuller:
Then we could just not load them in the first place, and load unlockables separately (in some resource package like challenges and mastery).

RDIL:
We'd also have to rework how plugins add items

AnthonyFuller:
That's fine, plugins are getting a re-work in v7 anyway.

PR #293, opened 2023-06-14
Amogus sus mk2

atampy25:
# Deez commit

### DIFF
--- a/components/controller.ts
+++ b/components/controller.ts
@@ -514,6 +514,12 @@ export class Controller {
                     lastServerSideData.contracts,
                 )) {
                     this.contracts.set(contractId, contractData)
+
+                    if (contractData.SMF.destinations?.addToDestinations) {
+                        controller.missionsInLocations[
+                            contractData.Metadata.Location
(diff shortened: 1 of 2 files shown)


components/controller.ts
```diff
@@ -514,6 +514,12 @@ export class Controller {
                     lastServerSideData.contracts,
                 )) {
                     this.contracts.set(contractId, contractData)
+
+                    if (contractData.SMF.destinations?.addToDestinations) {
+                        controller.missionsInLocations[
+                            contractData.Metadata.Location
+                        ].push(contractData.Metadata.Id)
+                    }
```
AnthonyFuller, line 522: Why not use `addBefore` and `addAfter` here? Seems it would be useful.

AnthonyFuller (changes requested):

components/controller.ts
atampy25: Amogus sus

AnthonyFuller (approved):

PR #294, opened 2023-06-14
Add SMF contract support (#293)

AnthonyFuller:
As title.


atampy25 (commented):
Does not LGTM, terrible code

PR #299, opened 2023-07-15
hotfix: fix call to getUnlockableById in planning

### LINKED ISSUE #298: Planning for limited-loadout locations is broken on v7
**Describe the bug**
When you try to access the planning page for missions taking place in Hokkaido or Miami, where you can't bring any items with you until you get to a mimimum mastery level, the server just errors out:
```
[11:06:47:884] [Debug | planningView] Looking up details for contract - Location:LOCATION_MIAMI (assembly:/_pro/scenes/missions/miami/scene_flamingo.entity)
[11:06:47:886] [Error | uncaught] Cannot read properties of undefined (reading 'size')
[11:06:47:886] [Error | uncaught] TypeError: Cannot read properties of undefined (reading 'size')
    at getUnlockableById (c:\Laragon\www\Peacock\components\inventory.ts:412:29)
    at planningView (c:\Laragon\www\Peacock\components\menus\planning.ts:381:35)
```

Here's my [DEBUG_PROFILE.zip](https://github.com/thepeacockproject/Peacock/files/12061093/DEBUG_PROFILE.zip) with the complete logs.

**To reproduce**
You can run Peacock from source using the latest commit from the `v7` branch. You can start any mission set in Hokkaido or Miami from the destinations or campains tab in the main hub.

**Additional context**
The code at fault inverted the parameters sent to the `getUnlockableById`. It expects an `id` and a `gameVersion` but the call from `components/menus/planning.ts` inverted the two, sending `gameVersion` first and then `id`: `{ id: 'h3', gameVersion: 'LOADOUT_UNLOCK_MIAMI' }`

What's weirder is that this code hasn't been touched since it was added, meaning this code never worked. And yet there's no problem in the Peacock release `v6.4.0`, even though the commit it belongs to has the bug:
```
const loadoutUnlockable = getUnlockableById(
    req.gameVersion,
    req.gameVersion === "h1"
        ? sublocation?.Properties?.NormalLoadoutUnlock[
              contractData.Metadata.Difficulty ?? "normal"
          ]
        : sublocation?.Properties?.NormalLoadoutUnlock,
)
```

This code should be included in all releases since `v6.1.0`, yet none of them have the issue, why doesn't the bug appear there? Unless the code is being patched, but the releases are built by CI it seems, and there's nothing there?

I feel like the solution is right under my nose and I'm just dumb, but I have a pull request ready to go that un-inverts the arguments sent to the function.

For reference, here's the link to the Discord thread in which I first reported the issue : https://discord.com/channels/826809653181808651/1129701652279070750

jorys-paulin:
The code in `components/menus/planning.ts` on line 381 calls the `` function with the wrong arguments, sending `gameVersion` then `id` where it should be `id` then `gameVersion`:
```typescript
const loadoutUnlockable = getUnlockableById(
    req.gameVersion,
    req.gameVersion === "h1"
        ? sublocation?.Properties?.NormalLoadoutUnlock[
              contractData.Metadata.Difficulty ?? "normal"
          ]
        : sublocation?.Properties?.NormalLoadoutUnlock,
)
```

This causes planning for Miami and Hokkaido, which use that code, to break. The server just errors out and the game is stuck until it times out.

This pull request fixes that inversion.

Fixes #298.

### DIFF
--- a/components/menus/planning.ts
+++ b/components/menus/planning.ts
@@ -379,12 +379,12 @@ export async function planningView(
         getFlag("enableMasteryProgression")
     ) {
         const loadoutUnlockable = getUnlockableById(
-            req.gameVersion,
             req.gameVersion === "h1"
-                ? sublocation?.Properties?.NormalLoadoutUnlock[
-                      contractData.Metadata.Difficulty ?? "normal"
-                  ]
-                : sublocation?.Properties?.NormalLoadoutUnlock,
+            ? sublocation?.Properties?.NormalLoadoutUnlock[
+                contractData.Metadata.Difficulty ?? "normal"
+            ]
+            : sublocation?.Properties?.NormalLoadoutUnlock,
+            req.gameVersion,
         )
 
         if (loadoutUnlockable) {
--- a/components/menus/planning.ts
+++ b/components/menus/planning.ts
@@ -380,10 +380,10 @@ export async function planningView(
     ) {
         const loadoutUnlockable = getUnlockableById(
             req.gameVersion === "h1"
-            ? sublocation?.Properties?.NormalLoadoutUnlock[
-                contractData.Metadata.Difficulty ?? "normal"
-            ]
-            : sublocation?.Properties?.NormalLoadoutUnlock,
+                ? sublocation?.Properties?.NormalLoadoutUnlock[
+                      contractData.Metadata.Difficulty ?? "normal"
+                  ]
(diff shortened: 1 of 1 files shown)


LennardF1989 (approved):
Thank you for your contribution! Classic case of swapped arguments. LGTM!

AnthonyFuller (approved):

PR #307, opened 2023-07-22
Added Dashes to Escalation names to be on par with ioi missions.

Cookie303728:
As the name says, the official german translation of iois missions have a dash between every escalation with "Name/Noun-Noun".
Oh and a single string i somehow missed.


VolkerSchlegel (approved):
LGTM

AnthonyFuller (approved):

PR #319, opened 2023-09-04
Fix to extension filtering in GetProfile request

vitphire:
It was a problem with no real effect, but a problem nontheless.
[Discussed in a bit more detail here.](https://discord.com/channels/826809653181808651/1089161116527108157/1148372825615388917)


AnthonyFuller (changes requested):
Please run prettier on your branch.

AnthonyFuller (approved):

PR #324, opened 2023-09-28
Fix SMF check

prolium:
This is an optional key, may not exist


AnthonyFuller:
Already did this on v6, guess I never propagated the change over.

PR #335, opened 2023-10-30
Refactor SMF support into it's own helper class

RDIL:
Provides a cleaner API surface, removes this code from just being in the controller.

!! BREAKING CHANGE !!

### DIFF
--- a/components/smfSupport.ts
+++ b/components/smfSupport.ts
@@ -1,0 +1,182 @@
+/*
+ *     The Peacock Project - a HITMAN server replacement.
+ *     Copyright (C) 2021-2023 The Peacock Project Team
+ *
+ *     This program is free software: you can redistribute it and/or modify
+ *     it under the terms of the GNU Affero General Public License as published by
+ *     the Free Software Foundation, either version 3 of the License, or
+ *     (at your option) any later version.
+ *
+ *     This program is distributed in the hope that it will be useful,
+ *     but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *     GNU Affero General Public License for more details.
+ *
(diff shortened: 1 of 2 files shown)


components/smfSupport.ts
```diff
@@ -0,0 +1,182 @@
...
+                    "lastDeploy.json",
+                )) ||
+            false
+        )
+    }
+
+    private async executePlugin(plugin: string) {
+        if (!existsSync(plugin)) return
+
+        await this.controller.executePlugin(
```
atampy25, line 60: Is `this.controller` actually defined? The constructor doesn't seem to set it.

components/smfSupport.ts
```diff
@@ -0,0 +1,182 @@
...
+            "boot",
+        )
+
+        const lastServerSideData = this.lastDeploy?.lastServerSideStates
+
+        this.handleUnlockables(lastServerSideData)
+        this.handleContracts(lastServerSideData)
+        this.handleBlobs(lastServerSideData)
+
+        if (lastServerSideData?.peacockPlugins) {
```
atampy25, line 163: The elvis operator here is not technically necessary since `lastServerSideStates` will always be defined if the framework has written something to AppData.

components/smfSupport.ts
RDIL: Syntax sugar, if you put access modifiers on a constructor parameter, it automatically adds this.x = x

components/smfSupport.ts
RDIL: The type can't guarantee that I think.

components/smfSupport.ts
atampy25: Oh cool, didn't know that

components/smfSupport.ts
atampy25: That's fine then

atampy25 (approved):

PR #340, opened 2023-11-05
Reworking of the config system

### LINKED ISSUE #339: Reworking of the config system
- [ ] Change the config system to lazy-load everything
- [ ] Use hooks to modify instead of just raw accesses

RDIL:
Fixes #339


components/configManager.ts
```diff
@@ -0,0 +1,452 @@
...
+ * @param clone If the value should be cloned (saves memory if false, but as a side effect, modifications will affect the actual config).
+ * @returns The config.
+ * @throws {Error} If the config file specified doesn't exist.
+ */
+export function getConfig<T = unknown>(config: ConfigKey, clone: boolean): T {
+    if (!Object.hasOwn(configs, config)) {
+        throw new Error(`Tried to lookup config that does not exist: ${config}`)
+    }
+
+    let c = configs[config]()
```
LennardF1989, line 403: So, we're fine this will modify the original object whenever clone is set to false? And if so, what will happen if a plugin gets a config that it already modified before? I'm aware this is how it worked originally, but we didn't have plugins modifying the false (yet), then, only core. And we are/were very careful to make sure we clone when we do that... Seems like a can of worms.

tests/src/oauthToken.test.ts
```diff
@@ -17,11 +17,11 @@
...
 /* eslint-disable @typescript-eslint/no-explicit-any */
-import { UserProfile } from "../../components/types/types"
-import { handleOauthToken, JWT_SECRET } from "../../components/oauthToken"
+import { UserProfile } from "../../components/types/types.ts"
+import { handleOauthToken, JWT_SECRET } from "../../components/oauthToken.ts"
 import { sign, verify } from "jsonwebtoken"
-import * as databaseHandler from "../../components/databaseHandler"
-import * as platformEntitlements from "../../components/platformEntitlements"
+import * as databaseHandler from "../../components/databaseHandler.ts"
+import * as platformEntitlements from "../../components/platformEntitlements.ts"
```
LennardF1989, line 24: Why do these need a .ts extension? Can't we fix that with a tsconfig.json? Haven't actually pulled yet to play with it, just wondering out loud.

LennardF1989 (changes requested):
See notes.

PS. The unit tests that use configs are broken now, we could have probably repurposed that config-mock.

tests/src/oauthToken.test.ts
RDIL: Basically had to switch the tests over to ESM because of this code right here which needs top-level await to ensure it's done the thing before the tests run but not import the module too early or it can't be mocked properly (I think) - I spent a lot of time getting it working, and it refused to do so without this, so I'm not complaining.

Screenshot 2023-11-06 at 18 24 05

components/configManager.ts
RDIL: This will modify the original object even if clone is true. It's a matter of do we want to always clone (very slow for big objects like the H1 templates) or only clone when we know the consumer code is going to modify the object being returned.

tests/src/oauthToken.test.ts
LennardF1989: Might have to play with the config a bit, I've had a Vite-based project setup that outputs ESM, but while I could still code in a NodeJS style. Will glance over it again sometime later this week. Maybe there is a better way to bootstrap the whole application for testing purposes...

components/configManager.ts
LennardF1989: I think in the end it's not really our responsibility. I do think that in the case of `clone == true`, we should tap the cloned object, not the source object. Since we always pass `clone` to the tap, it's up to the plugin to handle it gracefully.

I'm trying to think of situations where one might want to apply different logic based on `clone`, but can't really think of one right now. Maybe a non-issue?

RDIL: So I've been thinking on this, maybe we just have configs be modified by plugins when they are first (lazily) loaded? I can't think of any plugins that need to modify a config, say, after a specific server event is fired.

Issue #337, opened 2023-10-31
Extract routing logic from other logic


RDIL:
Tracking issue for extracting the routing/express logic from the actual server handling to make things just generally work better.

PR #344, opened 2023-11-25
Fix some strict type errors

RDIL:
More work towards making the codebase TS strict mode compliant.

Namely:
- Fixed some cases where TS couldn't narrow types well enough
- Added assertions for cases that shouldn't(?) happen
- Fixed some docstring issues
- Changed a few maps to objects just because we don't need to use maps to access well-known keys (e.g. game versions)

This does contain a few API changes:
- Some APIs now explicitly mark that they could return undefined (to match their actual behavior)
- `getLevelCount` now always returns 0 if given an undefined group contract, instead of throwing from accessing undefined keys.

A LOT of the remaining type issues actually boil down to the routers expecting RequestWithJwt when the express types don't guarantee that, which is very annoying. Blocked by #337

### DIFF
--- a/components/candle/challengeHelpers.ts
+++ b/components/candle/challengeHelpers.ts
@@ -111,13 +111,13 @@ export function inclusionDataCheck(
     if (!incData) return true
     if (!contract) return false
 
-    return (
+    return Boolean(
         incData.ContractIds?.includes(contract.Metadata.Id) ||
-        incData.ContractTypes?.includes(contract.Metadata.Type) ||
-        incData.Locations?.includes(contract.Metadata.Location) ||
-        contract.Metadata?.Gamemodes?.some((r) =>
-            incData.GameModes?.includes(r),
-        )
+            incData.ContractTypes?.includes(contract.Metadata.Type) ||
+            incData.Locations?.includes(contract.Metadata.Location) ||
+            contract.Metadata?.Gamemodes?.some((r) =>
+                incData.GameModes?.includes(r),
+            ),
     )
 }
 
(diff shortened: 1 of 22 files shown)


components/candle/masteryService.ts
```diff
@@ -301,12 +306,8 @@ export class MasteryService {
...
         locationParentId: string,
         gameVersion: GameVersion,
-    ): MasteryPackage {
-        if (!this.masteryPackages.get(gameVersion).has(locationParentId)) {
-            return undefined
-        }
-
-        return this.masteryPackages.get(gameVersion).get(locationParentId)
+    ): MasteryPackage | undefined {
+        return this.masteryPackages[gameVersion].get(locationParentId)
```
RDIL, line 310: `.get` will just return undefined, so no need to include a `.has`

components/utils.ts
```diff
@@ -338,6 +338,7 @@ function updateUserProfile(
                 {},
             )
 
+            // ts-expect-error Legacy property.
```
RDIL, line 341: Intentionally no `@`, since we don't yet have `strict` set to true, it causes an error because this is valid code when strict is false.

components/candle/challengeHelpers.ts
```diff
@@ -111,13 +111,13 @@ export function inclusionDataCheck(
     if (!incData) return true
     if (!contract) return false
 
-    return (
+    return Boolean(
```
AnthonyFuller, line 114: Why it couldn't infer this is beyond me. It's pretty obvious.

AnthonyFuller (approved):

components/candle/challengeHelpers.ts
RDIL: It thought the last item in the chain was returning the value, which it technically was, but that still could be cast to a boolean in conditions

components/candle/progressionService.ts
```diff
@@ -214,6 +218,11 @@ export class ProgressionService {
 
             // Update the EvergreenLevel with the latest Mastery Level
             if (isEvergreenContract) {
+                assert.ok(
+                    contract.Metadata.CpdId,
+                    "evergreen contract has no CPD",
+                )
```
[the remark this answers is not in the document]
RDIL: Yes

PR #352, opened 2023-12-15
Separate routing logic and business logic

### LINKED ISSUE #337: Extract routing logic from other logic

RDIL:
Fixes #337


LennardF1989:
I'll pull this in end-of-the-day and give it a checkup. Just to make sure I validate it accordingly: where does routing logic end in your opinion and where does business logic start? If I here this, I would say the implementation of endpoint should only ever call other functions, that do stuff, and the only thing the endpoint does is return that or decide on the appropriate statuscode. Meaning that endpoint implementations are very small and compherensible functions to eventually return a response.

Is that how you see it, too?

RDIL:
Yeah, that’s basically my philosophy. Endpoints only validate then call another function.

I tested this PR with Neko, all 3 game versions have no regressions.

components/types/types.ts
LennardF1989, line 1: We should probably agree on an "always `type` approach", it enforces us to keep the types updated with new attributes.

components/types/score.ts
```diff
@@ -103,7 +103,7 @@ export interface MissionEndEvergreenPayout {
     IsPrestige: boolean
 }
 
-export interface MissionEndResponse {
+export interface MissionEndResult {
```
LennardF1989, line 106: We should make a convention document that layout these things.

components/types/gameSchemas.ts
```diff
@@ -104,3 +104,51 @@ export type LoadSaveBody = Partial<{
     difficultyLevel: number
     contractId: string
 }>
+
+/**
+ * Query params that `/profiles/page/Safehouse` gets.
+ * Roughly the same as {@link SafehouseCategoryQuery} but this route is only for H1.
+ */
+export type SafehouseQuery = {
```
LennardF1989, line 112: I love these, we should do this for all endpoints. A separate query type (and a typed response preferably).

components/scoreHandler.ts
```diff
@@ -511,42 +510,45 @@ export function calculateSniperScore(
...
-    // TODO: For this entire function, add support for 2016 difficulties
-    // Resolve the contract session
-    if (!req.query.contractSessionId) {
-        res.status(400).end()
-        return
-    }
+export type MissionEndError = { errorCode: number; error: string }
 
-    const sessionDetails = contractSessions.get(req.query.contractSessionId)
+export async function getMissionEndData(
```
LennardF1989, line 515: This is a perfect example of how all endpoints should eventually work.

components/multiplayer/multiplayerService.ts
```diff
@@ -233,12 +233,12 @@ multiplayerRouter.post(
...
 )
 
-multiplayerRouter.post("/RegisterToPreset", jsonMiddleware(), (req, res) => {
+multiplayerRouter.post("/RegisterToPreset", jsonMiddleware(), (_, res) => {
     // matchmaking
     // TODO: implement matchmaking
     // req.body.presetId
     // req.body.lobbyId (this is just a timestamp?)
-    res.status(500).end()
+    res.status(501).end()
```
LennardF1989, line 241: Is this a thing the game now expects?

components/multiplayer/multiplayerService.ts
```diff
@@ -174,14 +174,14 @@ multiplayerRouter.post(
             // join existing match
             const match = activeMatches.get(req.body.matchId)!
 
-            match.Players.forEach((playerId) =>
+            for (const playerId of match.Players) {
```
LennardF1989, line 177: Is there a particular reason you did this, and if so, should we do it everywhere where forEach is used (about 26 times)? Should be added to convention list and potentially ban using eslint.

components/menus/playnext.ts
```diff
@@ -81,20 +88,20 @@ export function getSeasonId(index: number): string {
...
  * @returns The tile object.
  */
-export function createPlayNextTile(
+export function createPlayNextMission(
     userId: string,
     contractId: string,
     gameVersion: GameVersion,
     campaignInfo: PlayNextCampaignDetails,
-) {
+): PlayNextCategory {
```
LennardF1989, line 104: Note-to-self: Probably enable ESLint rule to force return type on functions.

components/menus/playnext.ts
```diff
@@ -21,11 +21,16 @@ import { generateUserCentric } from "../contracts/dataGen"
...
     MissionStory,
     PlayNextCampaignDetails,
+    UserCentricContract,
 } from "../types/types"
 
-export const orderedMissions: string[] = [
+/**
+ * Main story campaign ordered mission IDs.
+ */
+export const orderedMainCampaignMissions: string[] = [
```
LennardF1989, line 33: I feel we need like one file in the code base that has these kind of hardcoded lists, that way you will never had to scan the code and code only ever contains functions.

components/menuData.ts
```diff
@@ -1400,173 +781,112 @@ menuDataRouter.get("/LeaderboardsView", (req, res) => {
...
+        if ((planningData as PlanningError).error) {
+            res.status(400).send("check console")
+            return
         }
 
-        next("router")
+        const pd: GamePlanningData = planningData as GamePlanningData
+
+        // create contract planning isn't supposed to have the following properties
+        delete pd.ElusiveContractState
```
LennardF1989, line 873: Shouldn't we just make a new type, or map to a new one, rather than "clean up"?

components/configSwizzleManager.ts
```diff
@@ -280,7 +280,7 @@ export function getVersionedConfig<T = unknown>(
     if (
         // is this scpc, do we have a scpc config?
         gameVersion === "scpc" &&
-        Object.prototype.hasOwnProperty.call(configs, `Scpc${config}`)
+        Object.hasOwn(configs, `Scpc${config}`)
```
LennardF1989, line 283: Do we really need to do a `hasOwn` here? Does it matter if it is inherited, undefined check suffices? I feel this is a true for a lot of locations where we use hasOwn.

LennardF1989 (approved):
Just some questions here and there that need answering, other than that no objections.

Notes-to-self:
- Turn on strict mode with rules about usage of `!`
- Turn on additional ESLint rules to enforce unused/type-safety rules/return types
- Write down code conventions that are applicable to this project (not camelCase and such, but more like everything should be `type`, a response-model is always name `SomethingResult`, etc.
- Pull apart route/router definitions from their functions, separate files for both. All functions always have a return-types (both happy and unhappy if applicable) (force with ESLint).

components/multiplayer/multiplayerService.ts
RDIL: Foreach looks better imo

Issue #353, opened 2023-12-15
Proper killing conditions absent from displayed objective tips in certain peacock escalations


suanjiansalt:
**Describe the bug**
I've found that some peacock escalations have incorrect in-mission display texts that lack of killing conditions. For example:

- _The Christmas Calamity_ (southern comfort) : It says "Eliminate targets" but in this escalation you can just pacify them. Plus, this escalation is unique to lack of exclusive images, both for title and objectives.
- _The dez Dichotomy_ (rosebush) : Lack of disguise conditions (mansion staffs / guards). Plus, in Level 3 we need to eliminate the two Jay Harrises.
- _The mendietinha Madness_ (shangrila) : Lack of disguise condition (delivery man) + killing methods (Bartoli Woodsman Sniper Rifle, explosive accident).
-  _The Yannini Yearning_ (yannini) : Lack of killing methods (pistol).

I wanted to fix it myself, but struggled to find a way to perfect it. Mainly I don't even know exactly where to change.

suanjiansalt:
So finally I decide to dig into it...
The "BriefingName"&"BriefingText" lines are for sentences in planning section, and the "HUDTemplate" is for in-mission display sentence. The "ObjectiveType" determines icon before the text. All I need to do is replace "$loc localizations", right?
If I know that I would be able to start.

grappigegovert:
Only use iconType for the icon, ObjectiveType changes the way it's displayed in the planning page (see this function: https://github.com/thepeacockproject/Peacock/blob/v7/components/contracts/dataGen.ts#L354)
You can use `$($repository npc-repository-id-here).Name` to get an NPC's name, often used like this:
```json
"HUDTemplate": {
	"display": {
		"$loc": {
			"data": "$($repository f03a4c5a-767f-4315-a160-baef4bd1def0).Name",
			"key": "UI_CONTRACT_GENERAL_OBJ_KILL"
		}
	}
}
```
This uses the localization key "UI_CONTRACT_GENERAL_OBJ_KILL", which in english is "Kill {0}", and replaces the {0} with the NPC's name.

suanjiansalt:
> You can use `$($repository npc-repository-id-here).Name` to get an NPC's name, often used like this:
> 
> ```json
> "HUDTemplate": {
> 	"display": {
> 		"$loc": {
> 			"data": "$($repository f03a4c5a-767f-4315-a160-baef4bd1def0).Name",
[...]

PR #358, opened 2023-12-23
Fix peacock escalations' incomplete displayed killing tips

suanjiansalt:
Improvement program for issue #353 , huge thanks to @grappigegovert .

A dumb try, so please treat this as a pupil's homework, after all it's my first time doing this...

Note: 

- [x] Here I also did some adjustments to `The Christmas Calamity` gameplay. <s>If it's broken, I'm appreciated to receive your kind advice.</s> This time I fixed it correctly. If it's unnecessary, I will rollback it in no time.
- [x] Icon shown by `HUDTemplate` line looks improper in `The Tedious Thievery`. Before you collect all items (7 golds and 2 burgers) the icon is exactly the 'Exit' icon, but after completion the icon is what we want (a blue 'objective' icon). <s>I deleted `OnActive` line but this could be ridiculous. Please inform me before I look too stupid.</s> I changed the `iconType` to `16` and it looks good now.
- [x] Maybe all targets' name localization is required to realize?


contractdata/PARIS/ESCALATION/XMAS/XMAS1.json
```diff
@@ -76,7 +80,12 @@
                                             },
                                             {
                                                 "$inarray": {
-                                                    "in": "$.LethalWeapons",
+                                                    "in": {
+                                                        "$or": [
```
grappigegovert, line 84: Does this work?

AnthonyFuller (approved):
RSLGTM for l10n. A note: some lines that were previously translated to Chinese are no longer translated.

contractdata/PARIS/ESCALATION/XMAS/XMAS1.json
suanjiansalt: Sorry, but maybe I used too much imagination and creativity here. You may see that all I want to do is to take Christmas explosives into consideration, so maybe there's a more acceptable type exist?

grappigegovert: You could just add the two repo ids of the explosives to the `LethalWeapons` array. I'm not entirely sure if explosive weapons can be checked like this however, so I suggest testing it.

suanjiansalt: That makes sense... So the final question is, how to test? 🤯

grappigegovert: Play the contract and check if killing a target with your added explosives changes the counter?

suanjiansalt: I mean how to replace the game file with my JSON

suanjiansalt:
Decide to localize all killing tip strings. Sorry for the wavering.

contractdata/PARIS/ESCALATION/XMAS/XMAS1.json
grappigegovert: You can either follow the steps [in the Readme.md under "Installation"](https://github.com/thepeacockproject/Peacock?tab=readme-ov-file#installation), and just run a dev version of Peacock with your modified files, or if you want you could copy the contract json into the "contracts" folder of your release version of Peacock, and then access the contract through "My contracts" in game.

AnthonyFuller (approved):
I assume it all works and has been verified.

grappigegovert:
All of these "Eliminate <name>" lines without any condition don't really need dedicated localization and create unnecessary work for the translators, the generic UI_CONTRACT_GENERAL_OBJ_KILL should work fine

I've only really looked at the the first contract in the list, MENDIE1, and that one already was wrong: your new edit says that the kill needs to be an explosion accident, where any kill method is sufficient.
It also looks like you changed the locr key `UI_PEACOCK_SOUTHERN_COMFORT_OBJ1` to `UI_PEACOCK_SOUTHERN_COMFORT_XMASGIFT`, but in doing so changed the text for all languages to the english translation.

I'll do some more testing later

suanjiansalt:
> All of these "Eliminate <name>" lines without any condition don't really need dedicated localization and create unnecessary work for the translators, the generic UI_CONTRACT_GENERAL_OBJ_KILL should work fine
1. I did this because in H3, almost official escalations really localized all strings like "Eliminate <name>". For example, in _the Jinzhen Incident_:
Screenshot 2024-01-02 110035
Just eliminating Shihong Luo and Hui Hou in any conditions are both localized. Again, as I said earlier, if it's not necessary, just delete it.
> I've only really looked at the the first contract in the list, MENDIE1, and that one already was wrong: your new edit says that the kill needs to be an explosion accident, where any kill method is sufficient.
2. Oh YES, it is my fault. Not understand why I did it, but it clearly a fault I made on purpose.
> It also looks like you changed the locr key `UI_PEACOCK_SOUTHERN_COMFORT_OBJ1` to `UI_PEACOCK_SOUTHERN_COMFORT_XMASGIFT`, but in doing so changed the text for all languages to the english translation.
3. Well, I'm not merely replace the whole line with new key `UI_PEACOCK_SOUTHERN_COMFORT_XMASGIFT`. All I did in fact is change the keys' names, so I also give the right value to the keys if you look into it. This happens to some complications of _The Jeffery Confession_ too.

grappigegovert:
Alright, did a quick test with an xmax explosive, seems to work nicely.

The tileimage for the christmas calamity is broken with this change however, it would need to be updated in the image pack.

grappigegovert (approved):
Looks all good to me :)

RDIL:
Thanks @suanjiansalt!

Issue #333, opened 2023-10-27
Items not normally stashable can be equipped


RDIL:
**Describe the bug**
Seems certain items like muffins/crowbars can be stashed in small smuggle points, while the official servers do not allow this.

**To reproduce**
https://www.youtube.com/watch?v=53EuzPE56G0
Official servers don't allow bringing muffins to that stashpoint

**Additional context**

grappigegovert:
I think we might need to check the `ItemSize` property in the repository for this.

PR #359, opened 2024-01-06
Check for small item sizes for hidden stash

solderq35:
Fixes https://github.com/thepeacockproject/Peacock/issues/333

Used rpkg cli to check itemSize as suggested by @grappigegovert

NodeJS / RPKG CLI script I used: https://github.com/solderq35/hitman-tech-tips/blob/main/modding/add_itemsize.js

The only item in allunlockables.json that I couldn't find in rpkg was "PROP_MELEE_EIFFELSOUVENIR_CLUB" (repository ID: 7257eaa1-c8f3-4e0c-acbf-74f73869c1b2). So feel free to update the itemsize for that if it's known.

Edit: Also realized later that Brine Damaged SMG (repository ID: 79f8c0e9-4690-4ebf-b2b3-fd8411a1407f) has "ItemSize: small" despite being a "carriedweapon", so I changed the logic in menuData.ts to reflect that. The Brine Damaged SMG is equippable to hidden stash, but not concealed weapon slot on vanilla servers.

Use cases tested:
- Normal gear slot: everything except large weapons ("carriedweapon")
- Concealed weapon slot: pistols, dak x2 smg
- Inside briefcase: everything
- Agency pickup: everything
- Hidden Stash: Pistols, dak x2 smg, small gear / tools, small melee. Large or medium melee like muffins, crowbar are excluded

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,3 +56,6 @@ THIRDPARTYNOTICES.txt
 overrides
 
 DEBUG_PROFILE.zip
+
+packaging/add_itemsize/*
+!packaging/add_itemsize/add_itemsize.js
--- a/.prettierignore
+++ b/.prettierignore
@@ -10,3 +10,5 @@ components/contracts.json
 logs
 resources/**/*.json
 !resources/locale.json
+packaging/add_itemsize/*
+!packaging/add_itemsize/add_itemsize.js
--- a/components/2016/legacyMenuData.ts
+++ b/components/2016/legacyMenuData.ts
@@ -155,8 +155,8 @@ legacyMenuDataRouter.get(
                                             item.Unlockable.Properties
                                                 .LoadoutSlot !== "disguise")) && // => display all non-disguise items
                                     (req.query.allowlargeitems === "true" ||
-                                        item.Unlockable.Properties
-                                            .LoadoutSlot !== "carriedweapon") &&
+                                        item.Unlockable.Properties.ItemSize === // regular gear slot or hidden stash => small item
+                                            "ITEMSIZE_SMALL") &&
                                     item.Unlockable.Type !==
                                         "challengemultipler" &&
                                     !item.Unlockable.Properties.InclusionData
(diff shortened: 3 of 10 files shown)


RDIL:
This is awesome! Few concerns:

- this will break behavior for the other 3 supported games as they don’t have the relevant data
- how do we want to handle this for plugins

grappigegovert:
The allunlockables.json file used to be a straight up copy from the game, right?
That means we need to run this script to automatically add this data on every game update, and the script should thus also be in the repo I think.

My initial thought was to just include a copy of the full repository as well, but @RDIL will probably dislike adding that much data to memory.

While I'm not entirely sure how plugins would affect this, it will definitely not work for any items added with SMF, as the lastdeploy json currently only has a list of unlockables, and no repo entries.

RDIL:
The repository is a multi-megabyte game asset, and from a quick benchmark, it takes almost a second to parse and until it’s GC’d takes up ~20mb so I think shipping it is a bad idea

RDIL:
I think the best way to move forward here would be:

- add the script to the packaging folder
- restore the old logic for the other game versions? or do the same thing on both other games
- when we finish writing a plugin guide, we will note this as a caveat under items

solderq35:
I updated script / menudata logic to work for the other games (well I don't have Hitman sniper challenge, but I tested H1 / H2 / H3).

Also I'm pretty sure sniper challenge (Scpc) would be fine anyways, since the mode allows large weapon (sniper rifles) by default, and there are no agency pickup or hidden stash in that mode.

Usage instructions (also commented on the script):
 - Go to https://glaciermodding.org/rpkg/ and click "Download latest CLI"
 - Extract contents of downloaded zip file to `packaging/add_itemsize/`
 - From `packaging/add_itemsize/` directory, run: `node add_itemsize.js <path to Runtime> <Game Version (H1 / H2 / H3 / Scpc)>`
- E.g. `node add_itemsize.js "C:\Program Files\Epic Games\HITMAN3\Runtime" H3`
 - either H2 or H3 Runtime path seem to work for Scpc Game Version

Feel free to make any necessary adjustments to this.

I don't know enough about how SMF or plugins work to comment on that.

RDIL (approved):
Ready to merge upon approval from @grappigegovert for the new changes.

grappigegovert (approved):
Yep, looks good!

RDIL:
Thanks @solderq35!

PR #363, opened 2024-01-17
New non-pack challenges

moonysolari:
Some new non-pack challenges in 2023. Challenge packs is pending on config revamp and will come in a separate PR.


AnthonyFuller (approved):
Ahhhh finally. Thanks!

PR #369, opened 2024-01-22
Briefcase saved loadout fix

grappigegovert:
This should fix the error reported here: https://discord.com/channels/826809653181808651/1197928917290451014, where the planning page breaks because it tries to load a saved loadout with a briefcase that contains an item that is not in the user's inventory. (At least I think that was the issue, the user hasn't responded yet)
This can occur if you save a loadout, and then enable mastery progression or lose access to a dlc.

This targets v6 because I fear that v7 might not release for a while


RDIL (approved):
Looks good, thanks!

PR #373, opened 2024-01-24
Fix ArcadePageTemplate

### LINKED ISSUE #351: ETA menu does not work on v7
It works on v6 (including the new custom ETA) but not on v7. Can anyone reproduce?

image

grappigegovert:
Looks like a slight mistake was made while copy-pasting this switch case.
Fixes #351


moonysolari (approved):

RDIL (approved):
That's pretty funny. Thanks for the fix!

PR #370, opened 2024-01-22
Escalation QoL fix: Shangrila & Xmas

suanjiansalt:
The display problem in MendieMadness (#368 ) is a question caused by me, deeply confess here :(

As for Xmas disguise issue (#364 ), yeah I guess Mr. Reece might have not checked that meticulously yet, but I still want to submit it - It preliminarily fixes the two questions, which are banning disguise changes and making starting Santa 47 suit counts. However, I want it to go further. Can we enable disguise changes between Santa 47 suit and Santa 47 disguise? But for that I have no idea if we could and how we do.

fixes #368
fixes #364

PR #383, opened 2024-02-04
Fix the Christmas Calamity feat tile image

suanjiansalt:
Follow-up remedies of #370. (Maybe good at creating new problems by solving old ones?)


moonysolari (approved):
LGTM

PR #384, opened 2024-02-06
Fix some challenges not being ticked on mission end page

moonysolari:
If a challenge is not in the current location, it will not appear as "ticked" on the mission end page even if it is completed. This PR fixes that.


RDIL (approved):
Nice fix.

PR #391, opened 2024-02-19
Add test to ensure objectives are localized

RDIL:
Will avoid problems like #390 

Works by checking each contract's objectives's name, description, and HUD template to see if the strings are in either the game itself or our locale files.


moonysolari (approved):
LGTM

PR #394, opened 2024-03-03
Update locale.json

winnnty:
Added polish localisation. I'm not sure if this will work at all, because since Hitman 3, Polish is not officially supported.  However, there is a fan-made translation of the game and I tried to adapt to its standards.


RDIL (approved):
This will at least work in H2's engine, thank you for this!

PR #398, opened 2024-03-06
Add H2 ET challenges

suanjiansalt:
The Ids of these challenges have been compared and adjusted to those in the official server. (But there's no comparison for challenges like _The Forger_, _The Brothers_ that didn't appear in H2.) Also changed the path of the images to separate it from H3.

I can't fetch H2016 ET challenges with Fiddler so that's it...


suanjiansalt:
This has been implemented in v7.

PR #401, opened 2024-03-24
tests: Add new tests for user data persistence

RDIL:
Adds some new tests to make sure user data persistence works.


components/index.ts
```diff
@@ -629,7 +606,7 @@ program.option(
 )
 program.option(
     "--plugin-dev-host",
-    "activate plugin development features - requires plugin dev workspace setup",
+    "activate plugin development features - requires plugin dev workspace setupFileStructure",
```
AnthonyFuller, line 609: Find and replace moment

tests/src/__snapshots__/databaseHandler.test.ts.snap
AnthonyFuller, line 1: This isn't meant to be committed, right?

AnthonyFuller (changes requested):
Looks good! See comments.

components/index.ts
```diff
@@ -88,9 +87,7 @@ import { multiplayerRouter } from "./multiplayer/multiplayerService"
 import { multiplayerMenuDataRouter } from "./multiplayer/multiplayerMenuData"
 import { pack, unpack } from "msgpackr"
 import { liveSplitManager } from "./livesplit/liveSplitManager"
-import { cheapLoadUserData } from "./databaseHandler"
-
-loadFlags()
```
RDIL, line 93: somehow this ended up down here even though it's also called near the top?

components/index.ts
RDIL:
```suggestion
    "activate plugin development features - requires plugin dev workspace setup",
```

tests/src/__snapshots__/databaseHandler.test.ts.snap
RDIL: No, it is - it's a snapshot, meaning that tests have something to actually compare against, which does get committed because otherwise you'd have to generate the baseline by running against master and then the PR.

AnthonyFuller: Gotcha :+1:

AnthonyFuller (approved):
LGTM. Also: RIP Stadia.

RDIL:
Will fix the CodeQL issues before next release (they are actual exploitable issues this time!)

PR #402, opened 2024-03-24
escalation: yulania

suanjiansalt:
Another one.


AnthonyFuller (approved):
Fun escalation! Please PR images to the image pack then it will be merged.

suanjiansalt:
Emm... shall not we discuss more about the codename? Like changing it to Greek gods or something?

AnthonyFuller:
The codename is fine as it is, it's a convention we try to use, but as you have one already it's fine.

PR #405, opened 2024-03-30
Fix Evergreen and Sniper max level progression reset

Yanrishatum:
Upon reaching maximum level, the XP value gets clamped to maximum level XP value, and that caused entire for loop to fall through and return `1`, permanently resetting the player level to 1.
It now returns correct level when player reached the maximum.

Same bug was present in sniper map progression.


RDIL (approved):
Thanks for finding this!

Yanrishatum:
You may also want to cherry-pick it into v6, as it has the same bug afaik ;)

RDIL:
Good idea, will do that.

PR #407, opened 2024-04-01
Add challenge and restricted loadout objective for yulania

suanjiansalt:


AnthonyFuller (approved):
Nice catch before release!

PR #455, opened 2024-05-09
Properly respect SMF `peacockIntegration` key

atampy25:
No idea at what point that part of the code was written but the peacockIntegration check is in the wrong place


AnthonyFuller (dismissed):
Run prettier.

RDIL (approved):

PR #459, opened 2024-05-11
feat: Add API to simplify creation of global challenges

Yanrishatum:
## Scope

This PR aims to provide plguin creators a way to create global challenges that act along the lines of Arcade challenge group, where it shows all challenges regardless of location. 
Exact API is up to debate, and it's more to kickstart discussion on how to better implement it. Right now it's mirroring the way challenge packs are implemented.

Current API usage example:
```ts
controller.challengeService.globalGroups.set('group-key', {
    allLocations: true,
    gameVersions: ["h3"],
    groupId: 'group-key',
    location: "GLOBAL_GROUP_LOCATION"
})
```
With `allLocations` flag it will add the global challenge group location always, regardless if there are any challenges from that group linked to the location. (Use case: Roulette mod)
Otherwise it will do so only if location contains challenges from the same challenge group. (Example: Classic challenges)

## Test Plan

* [x] Finalize API (discussion)
* [x] Implement proper API
* [x] Test it
* [ ] Provide example plugin showcasing both additiong of custom challenges and creation of global challenges.

## Checklist

### DIFF
--- a/components/candle/README.md
+++ b/components/candle/README.md
@@ -12,6 +12,6 @@ Thanks to Moo for the name suggestion.
 
 Challenge data reside in files in `contractdata` ending in `_CHALLENGES.json`. When building, the `packResources` function in `buildTasks.mjs` will take all such files and pack them into files in the `resources/challenges` folder.
 
-Then, when the server starts, challenge data is loaded from the files in the `resources/challenges` folder into the memory, via the `_loadResources` function in `controller.ts`. It calls the `_handleChallengeResources` function that in turn calls `challengeService.registerGroup` and `challengeService.registerChallenge` to register the challenge group and the respective challenges.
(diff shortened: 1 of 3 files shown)


RDIL:
I like this! I'd be willing to merge if nobody has any other feedback on the API design.

Yanrishatum:
Changes from initial draft:
* `controller.challengeService.globalGroups` is now `controller.challengeService.globalMergeGroups` to make it clearer what it does.
* The `classic`, `elusive` and H1/H2 escalation challenges merges are now removed in favor of using new API (code simplification)
* Added `registerChallengeList` method, because regular register did a lot of needless work. Simplifies registration of multiple challenges at once as well as marginally speeds up server boot as it has to do less work when registering challenges.
* Added `hasGroup` method to check if group is registered.

The example plugin is ready, but I'm not sure where to put it.

RDIL:
I think the place for that would be on the WIP plugin development guide on the website

RDIL (approved):
LGTM!

PR #466, opened 2024-06-03
Print stacktrace for uncaught errors in plugins

grappigegovert:
There have been a few people trying to make plugins recently that are (I assume) beginning coders, posting some errors with their plugins.
Currently, we only print `Error while evaluating plugin`, and the error message, which is not super helpful.

This PR changes that to include a stack trace for the error.
Example:
example1

I thought about trimming the stack trace to only call sites in the plugin, but I'm not sure if that's a good idea.

### DIFF
--- a/components/controller.ts
+++ b/components/controller.ts
@@ -1106,8 +1106,13 @@ export class Controller {
 
             await (plugin as (controller: Controller) => Promise<void>)(this)
         } catch (e) {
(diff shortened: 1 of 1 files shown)


grappigegovert:
The stack trace looks slightly different when running the built version of course:
example2

AnthonyFuller:
I'm glad someone's finally got around to it!

RDIL:
Great change, thanks!

Issue #435, opened 2024-04-16
Challenges Unlocking Starting locations duplicate that start.


lazyRares:
**The Bug**
Completing a challenge that unlocks a starting location will duplicate that starting location if the user has already chosen to have all gear/starting locations unlocked.

**To reproduce**
Should be doable on any map that has a challenge that will unlock a starting location. But in my scenario I did this on Marrakesh with the Consulate Parking Garage starting spot, and the 'Floor It!' challenge.

1. Go To Marrakesh
2. (Optionally) Check Starting Locations
3. Complete 'Floor It!' 
4. Check Starting Locations, Duplicate of Consulate Parking Garage visible.

![image](https://github.com/thepeacockproject/Peacock/assets/33793448/f381918a-340e-42ff-b005-4fbbf90dd294)

RDIL:
Can you send a zipped copy of your userdata please? I want to see what’s going on with your unlockables.

lazyRares:
https://www.dropbox.com/scl/fi/rdh54kmq49m3r7xlndlor/userdata.zip?rlkey=imr1k8vydf9g69me6iv45da39&dl=0

grappigegovert:
Are unlockables even stored in userdata? Isn't this just grantDrops being called while it shouldn't?

AnthonyFuller:
Unlockables are not stored in userdata, they're dynamically generated. We should be able to safely return grantDrops after granting XP.

PR #481, opened 2024-06-26
Fix double-unlocking stuff when masteryprogression is disabled

grappigegovert:
Fixes https://github.com/thepeacockproject/Peacock/issues/435.


RDIL (approved):
🥳

PR #488, opened 2024-07-06
fix: hokkaido and haven default suit unlockable

### LINKED ISSUE #442: Kick Back Kimono challenge in Hokkaido not properly giving reward?
**Describing the bug**
In the title, doing that challenge in Hokkaido is supposed to unlock VIP Kimono as an outfit. Works on official server, for some reason this specific challenge doesn't work on Peacock

**To reproduce**
Do the challenge, it is ticked off, no outfit in inventory

**Additional context**
In my case it was progress transfer, but I don't think it is tied to the problem, as in Discord some person reported this long time ago, apparently having this problem even without any data transfers

image

### LINKED ISSUE #469: Unable to choose tactical wetsuit on Hawke's bay
**Describe the bug**
Impossible to select Tactical Wetsuit on Hawke's bay specifically. Even if it is unlocked by doing "Mills Reverie" escalation. Screenshots as an example

**To reproduce**
Just check Hawke's bay and go into outfit menu

**Additional context**
Screenshots by Westi
image
image

suanjiansalt:
## Scope

- [x] Fix #442 . But cannot figure out why Tactical Wetsuit can be unlocked but can only be accessible everywhere but in Hawke's.
- [ ] May close #469 . For reviewers: please check if the violating way is acceptable.

## Checklist

### DIFF
--- a/static/StashpointTemplate.json
+++ b/static/StashpointTemplate.json
@@ -435,12 +435,6 @@
                                                                                 },
                                                                                 "do": {
                                                                                     "$if ": {
-                                                                                        "$condition": {
-                                                                                            "$or": [
-                                                                                                "$not $eqs ($.Item.Unlockable.Id,TOKEN_OUTFIT_WET_SUIT)",
-                                                                                                "$not $eqs ($arg location,LOCATION_NEWZEALAND)"
-                                                                                            ]
-                                                                                        },
                                                                                         "$then": {
                                                                                             "$mergeobjects": [
                                                                                                 {
--- a/static/StashpointTemplate.json
+++ b/static/StashpointTemplate.json
@@ -435,6 +435,12 @@
                                                                                 },
                                                                                 "do": {
                                                                                     "$if ": {
+                                                                                        "$condition": {
+                                                                                            "$or": [
+                                                                                                "$not $eqs ($.Item.Unlockable.Id,TOKEN_OUTFIT_NEWZEALAND_HERO_NEWZEALANDSUIT)",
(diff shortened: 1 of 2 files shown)


Watchman3001:
If it somehow helps - wetsuit can technically be used in Hawkes', the first time you play it, or until you select some other suit. It is usable technically, just not in outfits selection

suanjiansalt:
> If it somehow helps - wetsuit can technically be used in Hawkes', the first time you play it, or until you select some other suit. It is usable technically, just not in outfits selection

It does, but because default loadout. We cannot manually pick the wetsuit up so here I just call it 'unaccessible'.

static/StashpointTemplate.json
```diff
@@ -435,12 +435,6 @@
                                                                                 },
                                                                                 "do": {
                                                                                     "$if ": {
-                                                                                        "$condition": {
-                                                                                            "$or": [
-                                                                                                "$not $eqs ($.Item.Unlockable.Id,TOKEN_OUTFIT_WET_SUIT)",
-                                                                                                "$not $eqs ($arg location,LOCATION_NEWZEALAND)"
-                                                                                            ]
-                                                                                        },
```
RDIL, line 443: I wonder why this code is in the template. In theory, there should be 2 wetsuits: one which is unlockable, and one which is the one that the map uses by default.

AnthonyFuller: Also this is not how you remove an if statement, I'm surprised this doesn't crash.

AnthonyFuller (changes requested):

RDIL (approved):

PR #494, opened 2024-07-15
Fix Part One users having access to Evergreen through the destination tile in the dashboard

thecatontheceiling:
## Scope

If you're using Part One with the latest version of Peacock (7.3.1) and you click on the Freelancer game mode tile in the dashboard, it lets you play Freelancer freely without any restrictions with all Hitman 2 and 3 locations working.
This fixes that. 

I just changed the tile to redirect to the locations page instead of the planning page, just like what pressing "Freelancer" in the game modes tab does.

I also removed the "auto-click" but I can add that back if the maintainers want to keep that.

## Test Plan

Made my changes, ran `yarn run-dev` and checked that the game wasn't letting me play Freelancer through the dashboard tile anymore.

## Checklist

### DIFF
--- a/static/EiderDashboard.json
+++ b/static/EiderDashboard.json
@@ -340,29 +340,10 @@
                 "actions": {
                     "accept": [
                         {
-                            "set-menu-context": {
-                                "value": {
-                                    "onpageopened": {
-                                        "set-selected": {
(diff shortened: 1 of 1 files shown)


AnthonyFuller:
Fixed in 9eb92f0.

thecatontheceiling:
much cleaner fix, nice

PR #512, opened 2024-08-18
Not load the Brothers unless corresponding mod installed

suanjiansalt:
## Scope

Remove the entry of The Brothers ET in H3 if the game doesn't deploy [the Brothers mod](https://www.nexusmods.com/hitman3/mods/375). Leaving it there at this condition is pointless.

## Checklist

-   [ ] I have verified my changes work

### DIFF
--- a/components/contracts/hitsCategoryService.ts
+++ b/components/contracts/hitsCategoryService.ts
@@ -148,7 +148,8 @@ export class HitsCategoryService {
         })
 
         // Elusives
-
+        const smfSupport = new SMFSupport(controller)
+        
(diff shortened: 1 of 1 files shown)


AnthonyFuller (changes requested):
The mod directly replaces the contract and removes the `Season` property meaning that the menu template would never show it. It is also available without any work in HITMAN 2016 and HITMAN 2.

AnthonyFuller:
Going to have to do a bigger change for this. Closing this PR, but it will be implemented.

PR #516, opened 2024-09-02
"Objective completed" bonus fix

grappigegovert:
Currently, gamechanger objectives are not considered when calculating whether or not to give out the "all objectives completed" bonus on mission end.
Escalations do not consider their objectives at all atm, because their group contract's objectives are checked instead (bug).

This pr fixes those two issues.

The bonus is still not entirely correct, as the `obj.IgnoreIfInactive && !isObjectiveActive` check uses session.completedObjectives, which is populated from `ObjectiveCompleted` events instead of from SMP output states,
and objectives that use `SuccessEvent` / `ResetEvent` / `FailedEvent` are not handled at all by the SMP.


RDIL (approved):
Wow this is much better, thanks!

PR #519, opened 2024-09-08
feat!: add game version to resolveContract

RDIL:
## Scope

Sorry Fusha in advance.
Migrates the contract resolution system to be versioned by game.

Includes backwards compatibility shims for plugins.

## Test Plan

I have not yet tested. 

## Checklist

-   [ ] I have verified my changes work

### DIFF
--- a/components/loggingInterop.ts
+++ b/components/loggingInterop.ts
@@ -21,14 +21,6 @@ import picocolors from "picocolors"
 import winston from "winston"
 import "winston-daily-rotate-file"
 
-const isDebug = ["*", "true", "peacock", "yes"].includes(
-    process.env.DEBUG || "false",
-)
-
-const isTest = ["*", "true", "peacock", "yes"].includes(
-    process.env.TEST || "false",
-)
-
 /**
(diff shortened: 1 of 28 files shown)


fu5ha (approved):
Chatting lgtm Chatting

(The stuff I know about aka livesplit stuff looks good to me)

components/loggingInterop.ts
```diff
@@ -62,16 +54,13 @@ export enum LogLevel {
...
-export enum LogCategory {
+export const enum LogCategory {
     /**
      * Remove the category from the log
      */
     NONE = "none",
-    /**
-     * Set the category to the name of the calling function
-     */
-    CALLER = "caller",
```
RDIL, line 74: Drive-by removal of this because it causes a deoptimization to occur for every call of log, meaning any function that calls log will be much slower as a result. For things like recursive or repeated calls, this causes exponential time loss.

fyi @LennardF1989 - future migration will be to add more manual category names. Eventually, we should probably make this required.

PR #526, opened 2024-10-01
fix: remove webstorm shill

juspky:
WebStorm will provide a better experience for development

928d8a9a57515320068a2c339e88fd28

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -36,13 +36,6 @@ then install the dependencies by using:
 The project is bundled with a server running configuration.
 Run that when editing the project with the IDE of choice, and it should open.
 
-We highly suggest you use [WebStorm](https://www.jetbrains.com/webstorm/) for
-development,
-but it is a paid product, so you may want to choose something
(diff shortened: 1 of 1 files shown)


RDIL:
slander, how could u

VolkerSchlegel:
Almost everyone of us uses WebStorm for development, we aren't forcing you to use it, if you want to use any other tool to make an actual PR feel free to do so, but that being said WebStorm is the tool the active devs have the most experience with and are more likely to be able to help with in case of setup issues (though that doesn't mean we won't try to help you if you don't use it), so recommending it is perfectly reasonable imo.

juspky:
@VolkerSchlegel My PR was merely a joke.

But honestly I think a project should be editor agnostic and I think its odd that a HITMAN server replacement readme voices opinions about any code editor.

You guys made a great piece of software btw.

RDIL:
It is editor agnostic, but we have more configurations in the repo for webstorm specifically, hence why we recommend it

PR #527, opened 2024-10-25
feat add Serverlist editor to the Patcher

Tondorian:
## Scope

Add a button to the patcher to open a new Dialog to edit the serverList
This pull request also added a dependency to Newtonsoft.Json
Also updated .NetFramework to 4.8 since 4.6 is EoL since April 22

## Test Plan

Clicked the new button
edit the entries and save them
changes are stored in "Servers.json"

## Checklist

-   [ ] I have run Prettier to reformat any changed files

### DIFF
--- a/patcher/MainForm.cs
+++ b/patcher/MainForm.cs
@@ -6,25 +6,19 @@ using System.Windows.Forms;
 using System.Drawing;
 using System.IO;
 using System.Text;
+using Newtonsoft.Json;
 
 namespace HitmanPatcher
 {
     public partial class MainForm : Form, ILoggingProvider
     {
         private static MainForm instance;
+        private GameServer[] gameServers;
 
         public static MainForm GetInstance()
         {
             return instance ??= new MainForm();
         }
-
-        private static readonly Dictionary<string, string> servers = new Dictionary<string, string>()
-        {
-            {"IOI Official", "config.hitman.io"},
-            {"Peacock Local", "127.0.0.1"}
-        };
-
-        private static readonly Dictionary<string, string> serversReverse = servers.ToDictionary(kvp => kvp.Value, kvp => kvp.Key);
 
         private void ToggleTheme(bool darkModeEnabled)
         {
@@ -60,6 +54,27 @@ namespace HitmanPatcher
         public MainForm()
         {
             InitializeComponent();
+            
+            // Try to get ServerList
+            try
+            {
+                var fileContent = File.ReadAllText("Servers.json", Encoding.UTF8);
+                gameServers = JsonConvert.DeserializeObject<GameServer[]>(fileContent);
+                if (gameServers == null || gameServers.Length == 0)
+                {
+                    gameServers = CreateDefautServerList();
+                }
+
+            }
+            catch ( FileNotFoundException ex)
+            {
+                gameServers = CreateDefautServerList();
+            }
+            catch (JsonSerializationException ex) 
+            {
+                gameServers = CreateDefautServerList();
+            }
+            
             logListView.Columns[0].Width = logListView.Width - 4 - SystemInformation.VerticalScrollBarWidth;
             Timer timer = new Timer
             {
@@ -90,6 +105,16 @@ namespace HitmanPatcher
                 ShowInTaskbar = false;
                 trayIcon.ShowBalloonTip(5000, "Peacock Patcher", "The Peacock Patcher has been started in the tray.", ToolTipIcon.Info);
             }
+        }
+
+        private static GameServer[] CreateDefautServerList()
+        {
+            GameServer[] gameServers = new GameServer[2];
+            gameServers[0] = new GameServer { ServerName = "IOI Official", ServerAddress = "config.hitman.io" };
+            gameServers[1] = new GameServer { ServerName = "Peacock Local", ServerAddress = "127.0.0.1" };
+            var serverList = JsonConvert.SerializeObject(gameServers);
+            File.WriteAllText("Servers.json", serverList, Encoding.UTF8);
+            return gameServers;
         }
 
         public void log(string msg)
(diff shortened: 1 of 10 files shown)


riisikumi:
Doesn't this already have an equivalent in the LocalGhost Patcher? I wonder why it hasn't been added to the Peacock one

grappigegovert (changes requested):
Noted some issues and other points.
Also, it would be preferable if this didn't use Newtonsoft.Json, as it would mean having to include a dll file along with the patcher exe.

LennardF1989:
I think we should first move to [patcher-core](https://github.com/thepeacockproject/Peacock/tree/dev/patcher-core). That is a more modern setup (UI separated from logic) and has support for DPI and Steam Deck (through CLI). And for size reasons, it can still back-compile to .NET 4.8. 

Also, couldn't we just get the actual serverlist editor from @grappigegovert version back in the source?

patcher/MainForm.cs
```diff
@@ -128,24 +153,41 @@ private void MainForm_FormClosing(object sender, FormClosingEventArgs e)
...
             }
 
             return hostname;
         }
 
         private void SetSelectedServerHostname(string input)
         {
-            if (!serversReverse.TryGetValue(input, out string result))
+            GameServer server = null;
+            if (input == "localhost" || input == "127.0.0.1")
```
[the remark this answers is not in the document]
Tondorian: I think I introduced this check for the edge case a user used a custom name for "Peacock Local" at the moment the program starts. At the moment it shows 127.0.0.1 or "Peacock Local"

LennardF1989:
> I used JSON because it is easy to read and understand and most typescript devs know it. but if an additional dll is not wanted I can use another format to

You mean C# right :P?

As for the replacement: https://www.nuget.org/packages/system.text.json - this is part of .NET (not .NET Framework) by default now. For simple situations, there is no need to use Newtonsoft anymore. Another reason why I feel we should first adopt the core version, then reapply this.

patcher/MainForm.cs
```diff
@@ -54,26 +55,14 @@ private void ToggleTheme(bool darkModeEnabled)
...
-            }
-            catch ( FileNotFoundException ex)
-            {
-                gameServers = CreateDefautServerList();
-            }
-            catch (JsonSerializationException ex) 
-            {
-                gameServers = CreateDefautServerList();
-            }
+            configLocation = $@"{appData}\PeacockProject\";
```
LennardF1989, line 64: Please always use `Path.Combine`.

patcher/MainForm.cs
```diff
@@ -91,6 +80,26 @@ public MainForm()
             {
                 CurrentSettings = new Settings();
             }
+            try
+            {
+                var fileContent = File.ReadAllText($@"{configLocation}\Servers.json", Encoding.UTF8);
```
LennardF1989, line 85: Same here.

patcher/ServerListEditor.cs
```diff
@@ -43,8 +43,13 @@ private void saveButton_Click(object sender, EventArgs e)
                 list.Add(new GameServer { ServerName = item.Cells[0].Value.ToString(), ServerAddress = item.Cells[1].Value.ToString() });
             }
             GameServers = list.ToArray();
+            string appData = Environment.GetFolderPath(Environment
+                .SpecialFolder
+                .ApplicationData);
+
+            var folder = $@"{appData}\PeacockProject\";
```
LennardF1989, line 50: Same here.

patcher/MainForm.cs
```diff
@@ -107,13 +116,14 @@ public MainForm()
...
-        private static GameServer[] CreateDefautServerList()
+        private GameServer[] CreateDefautServerList()
         {
             GameServer[] gameServers = new GameServer[2];
             gameServers[0] = new GameServer { ServerName = "IOI Official", ServerAddress = "config.hitman.io" };
             gameServers[1] = new GameServer { ServerName = "Peacock Local", ServerAddress = "127.0.0.1" };
             var serverList = JsonConvert.SerializeObject(gameServers);
-            File.WriteAllText("Servers.json", serverList, Encoding.UTF8);
+           
+            File.WriteAllText($@"{configLocation}Servers.json", serverList, Encoding.UTF8);
```
LennardF1989, line 126: Same here.

patcher/ServerListEditor.cs
```diff
@@ -43,8 +43,13 @@ private void saveButton_Click(object sender, EventArgs e)
...
             var serverString = JsonConvert.SerializeObject(GameServers);
-            File.WriteAllText("Servers.json", serverString, Encoding.UTF8);
+            File.WriteAllText(@$"{folder}Servers.json", serverString, Encoding.UTF8);
```
LennardF1989, line 52: Same here.

patcher/ServerListEditor.cs
LennardF1989: I also feel there should just be a single static readonly in a file with the path to the configuration and `server.json`.

Tondorian:
> You mean C# right :P?
I really mean Typescript because the core project is typescript :)
I was not aware of the [patcher-core](https://github.com/thepeacockproject/Peacock/tree/dev/patcher-core) branch before

> Also, please change the version to 4.6. This is pre-installed on all Windows 10 and up, preventing players from having to install .NET Framework.

I updated to 4.8 because Microsoft do not support 4.6 anymore, but will revert this change

patcher/MainForm.cs
Tondorian: I had stolen this line from `Settings.cs` :P

Tondorian:
> Also, please change the version to 4.6. This is pre-installed on all Windows 10 and up, preventing players from having to install .NET Framework. 
On my win 11 Pc 4.6 is not available to compile the project is 4.6.2 okay, too?
@LennardF1989

LennardF1989:
> > Also, please change the version to 4.6. This is pre-installed on all Windows 10 and up, preventing players from having to install .NET Framework.
> 
> On my win 11 Pc 4.6 is not available to compile the project is 4.6.2 okay, too? @LennardF1989

https://learn.microsoft.com/en-us/dotnet/framework/install/guide-for-developers

I had a reason for targeting 4.6.1 for the patcher-core. Pretty sure it was related to something that was simply unavailable in .NET 4.6 and/or more trouble than it's worth to backport. Can you target that? Get this: https://dotnet.microsoft.com/en-us/download/dotnet-framework/thank-you/net461-developer-pack-offline-installer

Tondorian:
I managed to find the target installer for 4.6 was a bit hidden

PR #532, opened 2024-11-18
Add toggle for Peacock-exclusive content

RDIL:
## Scope

Add toggle for Peacock exclusive content

Currently only works for challenges, as we don't really have a good way of telling what's a Peacock-exclusive contract, but that can be added as a follow-up.

## Test Plan

Peacock exclusive challenges do not appear, but others do

## Checklist

-   [X] I have verified my changes work

### DIFF
--- a/components/candle/challengeHelpers.ts
+++ b/components/candle/challengeHelpers.ts
@@ -28,6 +28,7 @@ import {
 import { SavedChallengeGroup } from "../types/challenges"
 import { controller } from "../controller"
 import { gameDifficulty, isSniperLocation } from "../utils"
+import { getFlag } from "../flags"
 
 /**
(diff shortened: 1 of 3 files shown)


suanjiansalt:
Some inventory sets are exclusive (non-vanilla) too - this PR is for exclusive *content* so I put it here.

Yanrishatum (approved):
Apart from the need to update a few challenges that do not bear the `peacock` tag despite being Peacock challenges (example is Rocco escalation challenge) - seems good to me.

AnthonyFuller (changes requested):
Blocking from merging, I have a plan.

suanjiansalt:
duplication of #644

PR #535, opened 2024-11-20
feat: season of drop challenge part 2

suanjiansalt:
## Scope

- Add 'Surgical Procedure' challenge.

## Test Plan

The challenge is good to go.

The strange order of feat challenges will be fixed soon.

## Checklist

### DIFF
--- a/contractdata/SAPIENZA/_SAPIENZA_CHALLENGES.json
+++ b/contractdata/SAPIENZA/_SAPIENZA_CHALLENGES.json
@@ -7600,6 +7600,96 @@
                     "Tags": ["story", "hard", "live", "feats"]
                 },
                 {
+                    "Id": "d7b1f3e5-5199-4f03-8b3c-13e2fde8463c",
+                    "Name": "UI_CHALLENGES_SAPIENZA_PLAGUEDOCTOR_NAME",
+                    "ImageName": "images/challenges/sapienza/sapienza_plaguedoctor.jpg",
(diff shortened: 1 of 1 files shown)


TheKotti:
Did the challenge work? I made this exact change locally and while the challenge showed up in the game, it would not unlock.

RDIL:
Just tried it and it worked for me 🤔

suanjiansalt:
> Just tried it and it worked for me 🤔

Just tried again and it didn't work for me... 😵

RDIL:
Rolled back pending further investigation

AnthonyFuller:
Works fine on official, does not work on Peacock (just tested). You have to exit for it to pop on official, and the in-game counter for guards killed does not work.

suanjiansalt:
The only attribution uncertain here is `$Value.ActorType`.

suanjiansalt:
Ah shite

grappigegovert:
> You have to exit for it to pop on official, and the in-game counter for guards killed does not work (although this is an official issue).

Oh god, is IO's statemachine parser on their server different from what's in the game?

RDIL:
@grappigegovert it's the same, but there may be some type casting that happens on the C# side

AnthonyFuller:
It's because they check if `1` is equal to `"1"` and the context listener is named incorrectly. Changing those fixes the challenge on Peacock and in-game. I don't think we should support `$eq: [1, "1"]` as this seems like a one-off. Pushing fixed challenge now.

PR #537, opened 2024-11-27
Some scoring fixes

grappigegovert:
It turns out that Peacock's "no noticed kill" bonus logic was wrong. It checked if a `MurderedBodyFound` event happened at the same time as a `Kill` event, which sounds logical, but that's not how IOI does it.

IOI just listens to the `Witnesses` event, which means that whenever you get "compromised", you lose the no noticed kill bonus.
Yes, that means that you can lose the bonus without even killing anybody. Thanks IOI.

Anyways, I've also fixed some edge-cases relating to 5-star no-sa scores, and added the "providence member eliminated" actionreward for carpathian, which was not in peacock yet?
In fact, the `NonTargetKillsAllowed` contract metadata typedef was even wrong? Not sure how that happened.

The "providence member eliminated" challenge is a bit odd to be in the global challenges I think, but I couldn't find a way to add it to the location challenge file without it appearing in the list of challenges in-game. (this is the only location-specific actionreward afaik)

### DIFF
--- a/components/eventHandler.ts
+++ b/components/eventHandler.ts
@@ -368,6 +368,7 @@ export function newSession(
             IsWinner: false,
             timerEnd: null,
         },
+        silentAssassinLost: false,
         challengeContexts: {},
     })
     userIdToTempSession.set(userId, sessionId)
@@ -989,6 +990,12 @@ function saveEvents(
                     session.bodiesFoundBy.add(
                         (<MurderedBodySeenC2SEvent>event).Value.Witness,
                     )
+
+                    if (
+                        !(<MurderedBodySeenC2SEvent>event).Value.IsWitnessTarget
+                    ) {
+                        session.silentAssassinLost = true
+                    }
                 }
 
                 break
--- a/components/eventHandler.ts
+++ b/components/eventHandler.ts
@@ -931,6 +931,7 @@ function saveEvents(
             case "Witnesses":
                 for (const actor of (event as WitnessesC2SEvent).Value) {
                     session.witnesses.add(actor)
+                    session.killsNoticedBy.add(actor)
                 }
 
                 break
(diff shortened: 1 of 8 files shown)


grappigegovert:
Ugh, all the tests failing of course

components/eventHandler.ts
```diff
@@ -989,10 +991,10 @@ function saveEvents(
                         (<MurderedBodySeenC2SEvent>event).Value.Witness,
                     )
 
-                    if (event.Timestamp === session.lastKill.timestamp) {
-                        session.killsNoticedBy.add(
-                            (<MurderedBodySeenC2SEvent>event).Value.Witness,
-                        )
+                    if (
+                        !(<MurderedBodySeenC2SEvent>event).Value.IsWitnessTarget
```
RDIL, line 995: I'm confused. What does this property actually check for?

Based on it's name, I could infer a few possible meanings:
- Is the witness a target?
- Has the target (kill) been witnessed

The only reason I ask is that in H2/3, you can commit an illegal action in front of a target, they spot you and become a witness, but then you kill them, and your SA comes back (as long as they didn't "tell anyone" - i.e. the orange doesn't spread).
This behavior is not the same in 2016, where if you are spotted, even if it's by your target, it's RIP SA.

This ties back to my original question, it's not clear what this property does. Can we add a docstring on the typedef with a brief explanation?

grappigegovert: The property is true if the witness is a target.
That means that with this PR, if any body is found by a non-target, you will lose SA and cannot regain it by killing that witness.
I'll add a docstring.

The 2016 logic should be unaffected by this change.
I think it is a bit clearer if you look at the changes per commit.

RDIL (approved):
LGTM, thanks for the clarification!

components/types/events.ts
```diff
@@ -153,6 +153,7 @@ export type PacifyC2SEvent = KillC2SEvent
 
 export type MurderedBodySeenC2SEvent = ClientToServerEvent<{
     Witness: string
+    /** `true` iff the witness is a target */
```
RDIL, line 156:
```suggestion
    /** `true` if the witness is a target */
```

grappigegovert: The double f was actually intentional here, but I guess it does look a little weird

Issue #21, opened 2022-11-13
Document and improve hooks


RDIL:
- [ ] Document the hooks
- [ ] Maybe improve the API?
- [ ] Add new hooks where plugin makers need them

AnthonyFuller:
Areas for hook improvements:
- Challenges
- CPD (Maybe allow modders to do some cool stuff with this? May not need hooks.)
- Mastery
- Scoring(?)
- Multiplayer(?)

RDIL:
Additionally, add ability to untap, add priorities, etc

AnthonyFuller:
Can we not untap? Thought we could.

RDIL:
I don't think we can officially right now, you can always hack around it e.g.

```ts
let a = true

hooks.example.tap("MyPlugin", () => {
    if (!a) return
    console.log("still tapped")
    a = false
    console.log("untapped for future use")
})
```

AnthonyFuller:
Ah okay, that's fair. Should we still use tapable or move away from that? I vaguely recall you complaining about it in the past.

RDIL:
We already migrated off tapable because it's codegen at runtime approach could be buggy, [this](https://github.com/thepeacockproject/Peacock/blob/v7/components/hooksImpl.ts) is what we're using rn

AnthonyFuller:
Ah right I forgot about that change, mostly because we kept the API design the same.

PR #543, opened 2024-12-21
Add/fix some plugin hooks

grappigegovert:
This PR adds some small things useful for plugins that I was missing:
 - Adds an extra hook `onUserLogin` that gets called when a user logs in (requests `/configuration/Init`);
 - Adds back functionality to the `onMissionEnd` hook that already existed but was never called;
 - Passes a plugin's filename to the plugin, which can especially be useful for SMF-added plugins. (Basically just https://nodejs.org/api/modules.html#__filename)
 
Relevant issue: #21

### DIFF
--- a/components/controller.ts
+++ b/components/controller.ts
@@ -1122,6 +1122,7 @@ export class Controller {
             process,
             fetch,
             require: createPeacockRequire(pluginName),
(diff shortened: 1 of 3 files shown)


RDIL (approved):
Thanks!

PR #544, opened 2024-12-23
feat(controller): load contracts asynchronously

shadow578:
## Scope

this PR adds a small change so that locally saved contracts (in ./contracts/) don't block the server start.
the server will start immediately, and continue to load contracts in the background.

## Test Plan

connecting to the server should work quickly, even with a lot of contracts in the local folder.

## Checklist

-   [X] I have verified my changes work

### DIFF
--- a/components/controller.ts
+++ b/components/controller.ts
@@ -479,7 +479,11 @@ export class Controller {
 
         this._addElusiveTargets()
         this._getETALocations()
-        await this.index()
+
+        // load contracts asynchronously to avoid blocking the server
+        // otherwise, the player may have to wait a long time for the
+        // server to start, even if they're not planning to play contracts
+        this.index()
 
         try {
             await this._loadResources()
@@ -857,6 +861,8 @@ export class Controller {
                 log(LogLevel.DEBUG, e, "contracts")
             }
         }
+
+        log(LogLevel.INFO, "Contracts loading completed", "contracts")
     }
 
     /**


components/controller.ts
```diff
@@ -857,6 +861,8 @@ export class Controller {
                 log(LogLevel.DEBUG, e, "contracts")
             }
         }
+
+        log(LogLevel.INFO, "Contracts loading completed", "contracts")
```
RDIL, line 865: Maybe it would be best to do this as a `.then` on the Promise that `index` returns so that it's more accurate?

RDIL (commented):
Thanks for this! Some feedback

components/controller.ts
AnthonyFuller: Didn't notice that, good point. I'll add edit it now.

shadow578: This shouldn't matter, no? In both cases, the log is called after all promises are awaited

shadow578:
> I assume you noticed this after loading the entire contracts repo for your "offline" (aka piracy) branch.

Well, wanting to play offline doesn't mean I'm doing anything nefarious. Assuming such things imo isn't a great way to start a conversation.

But if you really think it could be used for such a purpose, I'd be ok to remove the branch.

AnthonyFuller:
What do you mean? You return a list of all entitlements no matter what the user owns. That is textbook piracy.

components/controller.ts
RDIL: The other promises are blocked on their awaits, but just because they are guaranteed to be done before the first one doesn't mean the first one isn't.

AnthonyFuller:
It's much appreciated, the proper way to do it would be making the user connect online first to grab their entitlements. Then using the stored versions in userdata when offline.

PR #550, opened 2025-01-04
refactor!: game versioned missionsInLocation, fix certain contracts when resolved, rename missionsInLocations -> missionsInLocation

AnthonyFuller:
Fixes #510 #505 #504

Needs testing and checking over more than twice.

## Breaking Changes
- `controller.missionsInLocations` is now `controller.missionsInLocation` (as it always should have been!).
- The aforementioned `controller.missionsInLocations` is now **split by game version**.
- `controller.addEscalation` has a new parameter: `(groupContract, locationId, **gameVersion**, ...levels)` (highlighted using asterisks).

---

Also this removes some terrible hacks we've had to do. We should've always done it like this!

PR #562, opened 2025-02-09
feat: add deprecation warnings to old addEscalation/missionsInLocations

grappigegovert:
The changes in v8 for missionsInLocation and addEscalation from #550 are breaking a bunch of plugins currently.
This PR adds back functionality for the old usages with a deprecation warning, to give plugin devs time to update.

### DIFF
--- a/components/controller.ts
+++ b/components/controller.ts
@@ -364,6 +364,21 @@ export class Controller {
     }
     public missionsInLocation = missionsInLocation
     /**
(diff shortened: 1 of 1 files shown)


RDIL (approved):

RDIL:
Agree that since this is a pretty critical API we should give at least 1 major for updates

AnthonyFuller (approved):
I was originally against this since we're going into a major version release, but I suppose it makes sense.

PR #569, opened 2025-03-10
feat: add priority to hooks and a blackboard for plugins

AnthonyFuller:
- Adds a priority option to taps and intercepts (intercepts may be removed)
- Adds a blackboard object (`controller.blackboard`) for plugins to share data

### DIFF
--- a/components/controller.ts
+++ b/components/controller.ts
@@ -380,6 +380,9 @@ export class Controller {
     public locationsWithETA = new Set<string>()
     public parentsWithETA = new Set<string>()
 
+    // A place for plugins to share data.
+    public blackboard: unknown = {}
+
     /**
      * The constructor.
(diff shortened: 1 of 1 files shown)


AnthonyFuller:
Ah intercepts were removed on master,

components/controller.ts
```diff
@@ -380,6 +380,9 @@ export class Controller {
     public locationsWithETA = new Set<string>()
     public parentsWithETA = new Set<string>()
 
+    // A place for plugins to share data.
+    public blackboard: unknown = {}
```
RDIL, line 384: Record<string, unknown> so users don't have to cast the parent, only the member being requested

RDIL (changes requested):

PR #570, opened 2025-03-10
Added support for running node server in Docker

m0ngr31:
## Scope

This is the initial work to allow a user to run the server in a Docker/Podman container. This can be automated with Github Actions to deploy images to Docker Hub or GHCR automatically on release.

Would be a big help for those of us who run all self-hosted software on a dedicated server.

Once this is approved, it'd be willing to help with adding instructions so that it's easy to run for people. I can also help getting the Github actions setup (for autopublishing and multi-arch build).

## Test Plan

I've run this locally and it seemed to work okay. I just discovered this project yesterday so I'll need to put it through its paces more

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,5 @@ DEBUG_PROFILE.zip
 packaging/add_itemsize/*
 !packaging/add_itemsize/add_itemsize.js
 /.idea/AIAssistantCustomInstructionsStorage.xml
+
+config/
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+FROM alpine:3.21.3
+
+RUN apk add --update nodejs npm su-exec shadow
+
+RUN rm -rf /var/cache/apk/*
+
+RUN mkdir /app_build
+WORKDIR /app_build
+RUN npm i -g yarn
+
+COPY . .
+
+RUN yarn install && yarn run build
+
+RUN chmod +x entrypoint.sh
+
+ENTRYPOINT ["./entrypoint.sh"]
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -1,0 +1,20 @@
+#!/bin/sh
+
+mkdir -p /app
+
+cp /app_build/chunk0* /app
+cp -r /app_build/resources /app
+cp -r /app_build/webui /app
+
+cd /app
+
+if [ -z "$PUID" ] || [ -z "$PGID" ]; then
+  exec node chunk0.js
+else
+  adduser -u $PUID -D abc
+  groupmod -g $PGID abc
+
+  chown abc:abc -R /app
+
+  su-exec abc node chunk0.js
+fi


Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM alpine:3.21.3
```
RDIL, line 1: May as well just use `node:22-slim`? Unless there's a specific reason for using alpine directly.

RDIL (commented):
Thanks for this!! Been meaning to add a Docker setup for a while.

AnthonyFuller (changes requested):
Personally, I think it'd be great to have a Docker container for Peacock. I have a few issues with what you've done:
- Installing node from a package manager does not guarantee that it is our required node version (see `.nvmrc`).
- It should not be doing a build step, it should be using the linux version of a release (needs to be a known build and then we can release containers for each version).
- There is no way to configure `options.ini` or to control the volume (granted this partly requires some Peacock changes, which now that I think about them, are not a bad idea).

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+RUN apk add --update nodejs npm su-exec shadow
+
+RUN rm -rf /var/cache/apk/*
+
+RUN mkdir /app_build
+WORKDIR /app_build
+RUN npm i -g yarn
```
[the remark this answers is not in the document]
m0ngr31: The idea of this was to copy what was needed into volume being passed in at runtime for persistent data. So you'd pass in like: `-v ./config:/app` as part of your Docker start command.

Ideally we could update the server with an environment variable or something so that persistent data could go into a custom directory

m0ngr31:
>     * Installing node from a package manager does not guarantee that it is our required node version (see `.nvmrc`).

You're totally correct about this. I will modify that.

>     * It should not be doing a build step, it should be using the linux version of a release (needs to be a known build and then we can release containers for each version).

If this is automatically run on an action as part of a release, wouldn't it end up being the same either way? I can try and familiarize myself with your actions and see if I can make it work like this from the build artifacts?

>     * There is no way to configure `options.ini` or to control the volume (granted this partly requires some Peacock changes, which now that I think about them, are not a bad idea).

That's where documentation would come in. With the current setup, it's building everything in `/app_build` and then at runtime it's copying that into `/app` which would be mounted as a volume when you're running the container. If the persistent data could be configured to a different directory we could skip that step

Dockerfile
m0ngr31: I updated this, good callout

m0ngr31:
@AnthonyFuller @RDIL 

Okay, I've updated the container build to use the artifact created by Cirrus CI. It'd need some tweaking based on your credentials for Docker Hub/GHCR and to make sure it is able to build the container inside of another container. I don't have access to your CI, so I might need some help.

But otherwise - this container would allow a user to have a persistent volume and keep using whatever version the container builds from

m0ngr31:
Okay, after 30 commits I finally got it working to build the container in Cirrus. I'll need your help with getting the vars setup but then it should be good

.gitignore
```diff
@@ -61,3 +61,5 @@ DEBUG_PROFILE.zip
 packaging/add_itemsize/*
 !packaging/add_itemsize/add_itemsize.js
 /.idea/AIAssistantCustomInstructionsStorage.xml
+
+config/
```
RDIL, line 65: what's this for?

m0ngr31: Oh that's a directory I had made for mounting the container in while testing. I can remove that line if needed

PR #572, opened 2025-03-17
fix: New York uniquely requires 4500XP per level instead of 6000

Yanrishatum:
## Scope

As we discovered, official servers require only 4500 XP for New York per level in H3, making it possible to reach level 20 with challenges it has (it's impossible with 6k per level).
Updates xp calculation to accept optional xpPerLevel argument, used exclusively by New York. Mastery data now has `XpPerLevel` nullable field similar to `MaxLevel`.

## Test Plan

Go to New York, get levels. Level up at 4500XP.

## Checklist


RDIL (approved):
LGTM, just make sure to run prettier

PR #582, opened 2025-05-08
Updated current patcher

LennardF1989:
Applied some changes from #388:
- Switched to SDK-style, but still builds as net46.
- Separated patcher logic from UI

This allows me to rebase my dev/patcher-core branch later and make the actual changes compared to the net46 version a lot easier to review/check.

Draft PR so I can test CI/CD while I'm at it.

This PR will get the "Re-enable Dynamic Resources"-patch before switching to a normal PR for consideration.


LennardF1989:
Last commit adds patch to re-enable Dynamic Resources starting 3.210 (technically untested, since I'm already on a slightly higher version). See [this](https://github.com/thepeacockproject/Peacock/pull/582/commits/2d2050971a658c3ae709801e0a307932adfcdb94) for easy comparison. 

The patch is treated as optional, meaning it does not fail a offset scan if it's not found. I kept the newly added code in line with however the original is structured, since my intention is not to do a refactor.

You can download the build from the CI [here](https://github.com/thepeacockproject/Peacock/actions/runs/14918296841/artifacts/3090417447). 

Acceptance criteria:
- Should be more or less the same size as the original PeacockPatcher.exe that's in the root of the project
- Should work with existing configuration (peacock_patcher.conf)
- Should restore loading Dynamic Resources, mostly visible by going to the main menu in-game. You should **not** see [UI_PEACOCK...] anymore.

patcher/HitmanPatcher.Core/AOBScanner.cs
```diff
@@ -96,7 +105,8 @@ public static bool TryGetHitmanVersionByScanning(Process process,
             Note("AuthHeader2", getAuthheadPatches.Result.First()[1]);
             Note("ConfigDomain", getConfigdomainPatches.Result.First()[0]);
             Note("Protocol", getProtocolPatches.Result.First()[0]);
-            Note("DynamicResources", getDynresForceofflinePatches.Result.First()[0]);
+            Note("DynamicResources->ForceOffline", getDynresForceofflinePatches.Result.First()[0]);
```
RDIL, line 108: I was going to say something about this potentially being renamed to make it clear that it's changing it to be not mandatory, but on second thought I think this is fine as-is. just noting for posterity

RDIL (approved):
LGTM!

patcher/HitmanPatcher/HitmanPatcher.csproj
```diff
@@ -0,0 +1,61 @@
...
+        <RootNamespace>HitmanPatcher</RootNamespace>
+        <AssemblyName>PeacockPatcher</AssemblyName>
+
+        <Title>Peacock Patcher</Title>
+        <Description>Peacock's HITMAN™ World of Assassination trilogy game patcher.</Description>
+        <Company>The Peacock Project</Company>
+        <Product>Peacock Patcher</Product>
+        <Copyright>Copyright © 2020-2025 grappigegovert &amp; The Peacock Project</Copyright>
+        <AssemblyVersion>8.0.0.0</AssemblyVersion>
+        <FileVersion>8.0.0.0</FileVersion>
```
RDIL, line 20: let's bump the major version

patcher/HitmanPatcher.Core/AOBScanner.cs
AnthonyFuller: Would be better to rename the notes to be the name of the original config commands (where applicable).

LennardF1989: I have been very tempted to do that (rename it to OnlineResources), but I wanted to go as-is and stay in line with what is already there. 

With this PR out of the way I intend to do a refactor with the .NET Core version and cleanup the code. I'd say that's a good time to also name things the way they should be. There are issues open about the CLI arguments/options being negatively named (noforceoffline), the fact some options should not even be an option, etc.

I therefor propose to leave the renaming out of this PR.

RDIL: SGTM

Issue #436, opened 2024-04-18
Automate patcher binary updates


RDIL:
We should upload patcher binaries from CI, and ideally make them fully reproducible. Committing the pre-built binary presents a security risk, as it is not clear if it has been tampered with.

RDIL:
Pushing directly seems fine. The historical reason for committing the binary to Git is so people working on the code can get up and running immediately. I agree we probably don't need to try and update it after every commit.

PR #583, opened 2025-05-09
Added manual workflow to publish patcher

LennardF1989:
PR for #436. 

Draft to test CI.

### DIFF
--- a/.github/workflows/patcher-publish.yml
+++ b/.github/workflows/patcher-publish.yml
@@ -1,0 +1,49 @@
+name: Patcher - Publish
+
+on:
+  workflow_dispatch:
+
+env:
+  GH_TOKEN: ${{ github.token }}
+
+jobs:
+  publish:
+    name: Publish Patcher
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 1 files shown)


LennardF1989:
This should do @RDIL!

While it may seem a little convoluted, I purposely kept the build steps separate, especially if we need to build multiple platforms later. Technically it could be two steps ([checkout], and [download, copy, commit, push]).

See https://github.com/thepeacockproject/Peacock/actions/runs/14933260137/job/41954609337

.github/workflows/patcher-publish.yml
```diff
@@ -0,0 +1,52 @@
...
+  GH_TOKEN: ${{ github.token }}
+
+jobs:
+  publish:
+    name: Publish Patcher
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
RDIL, line 19: Pin to exact hash for security

RDIL (approved):
LGTM with nit

PR #590, opened 2025-06-07
fix: sort contracts for deterministic build

NyCodeGHG:
## Scope

While packaging peacock using the [Nix package manager](https://nixos.org), I noticed that the build process is not reproducible because `fast-glob`'s order is not deterministic.

## Test Plan

I've rebuilt the project cleanly multiple times to make sure it's reproducible now (which it is ^-^!)

## Checklist

### DIFF
--- a/packaging/buildTasks.mjs
+++ b/packaging/buildTasks.mjs
@@ -124,6 +124,7 @@ export async function packResources() {
     const start = Date.now()
 
     const contracts = await glob("contractdata/**/*.json")
+    contracts.sort()
     const b = []
     const el = []
 


packaging/buildTasks.mjs
```diff
@@ -124,6 +124,7 @@ export async function packResources() {
     const start = Date.now()
 
     const contracts = await glob("contractdata/**/*.json")
+    contracts.sort()
```
RDIL, line 127: Thanks for the PR! I'm not sure if this actually addresses the root issue though, as `.sort` will just be calling `toString` on each object, which should theoretically be `[object Object]`. Therefore, I think we should explicitly define a sorting factor here, something like `contractData?.Metadata?.Id`

NyCodeGHG: `contracts` is a string array, this just sorts the files alphabetically

RDIL (approved):

PR #604, opened 2025-07-08
Fix freelancer mastery inventory items unlock

hdr88:
## Scope

This reverts change in a85f1d2.
It makes freelancer mastery inventory items locked if the necessary mastery level is not reached.
Currently all freelancer mastery inventory items are unlocked even if mastery level is not reached.

## Test Plan

Checked inventory after the change, only the items are there now that i have unlocked in freelancer mastery.

## Checklist

- [X] I have run Prettier to reformat any changed files
- [X] I have verified my changes work

### DIFF
--- a/components/inventory.ts
+++ b/components/inventory.ts
@@ -128,7 +128,9 @@ function filterUnlockedContent(
 
         // Handles unlockables that belong to a package or unlocked gear from evergreen
         if (packagedUnlocks.has(unlockable.Id)) {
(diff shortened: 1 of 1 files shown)


RDIL (commented):
I'm curious as to why exactly this fixes it 🤔

Looking at the code, `Map#get` should return undefined, which is falsey, so I'm not sure why this second check is needed. I'm probably missing something here

grappigegovert:
Looks good, thanks for the fix!

RDIL:
Thanks @grappigegovert, makes sense.

PR #605, opened 2025-07-13
feat: localize ingame options

suanjiansalt:
## Scope

Localize ingame options.

## Test Plan

Visual effect works.

## Checklist


RDIL:
I think this is a good change, but we'll want to accompany it with some way to build dynamic resources on the fly as a plugin API because then plugins can specify localization as well

PR #638, opened 2025-12-01
feat: support LocationSuitOverrides and refactor loadout logic so it is consistent

AnthonyFuller:
As title, this was hell.

### DIFF
--- a/components/loadouts.ts
+++ b/components/loadouts.ts
@@ -198,9 +198,8 @@ export class Loadouts {
 
         if (getFlag("loadoutSaving") === "LEGACY") {
             loadout =
-                userProfile.Extensions.defaultloadout?.[
-                    sublocation?.Id || ""
-                ] ?? defaultLoadout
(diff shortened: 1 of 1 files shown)


components/inventory.ts
```diff
@@ -547,22 +547,24 @@ export function getUnlockablesById(
...
+    suitOverride?: string,
 ): InventoryItem[] {
     if (!sublocation) {
         return inv
     }
 
     // Yes this is slow. We should organize the unlockables into a { [Id: string]: Unlockable } map.
     const locationSuit = getUnlockableById(
-        getDefaultSuitFor(sublocation),
+        getDefaultSuitFor(sublocation, suitOverride),
```
grappigegovert: I think it would be better if this were something like `suitOverride && getUnlockableById(suitOverride) ?? getUnlockableById(getDefaultSuitFor(sublocation))` so we can fall back to the location default for contracts with invalid values here

AnthonyFuller:
GitHub what are you doing, I MERGED the branch, christ.

PR #641, opened 2025-12-02
Remove XP bonus for NY exclusive challenge

RDIL:
## Scope

@thekotti & @Yanrishatum requested this, and better late then never

## Test Plan

Confirm no XP is granted when the challenge is completed

## Checklist

### DIFF
--- a/contractdata/MIAMI/_MIAMI_CHALLENGES.json
+++ b/contractdata/MIAMI/_MIAMI_CHALLENGES.json
@@ -3212,7 +3212,7 @@
                     "ImageName": "images/challenges/miami/dolphin.jpg",
                     "Description": "UI_CHALLENGES_SAPIENZA_EXP_KRAKEN_DESC",
                     "Rewards": {
-                        "MasteryXP": 1000
+                        "MasteryXP": 0
                     },
                     "Drops": ["TOKEN_OUTFIT_HERO_BLUESPECIAL_SUIT"],
                     "IsPlayable": true,


suanjiansalt:
I wonder why is Call of Blue not included...?

RDIL:
Oh? which map is that on

suanjiansalt:
Miami dolphin exit one

grappigegovert:
It would be nice if all peacock-exclusive challenges at least had a property to indicate them as such

Yanrishatum:
> A flag to filter "peacock" tagged challenges could meet your demand, I think?

Exactly what I would want, but it does require a fair bit of legwork in how those challenges are initialized.

> It would be nice if all peacock-exclusive challenges at least had a property to indicate them as such

Internally, at least, they have a tag `peacock` on those challenges (Except Butcher's Delight pack for some reason, but it does have it's own `cheesecake-pack` tag). While mods like heritage-ui can leverage that to provide information to user, it is clientside. Serverside it could be done with putting a note in the description or alternatively a peacock icon on the challenge tile.

TheKotti:
> Not much we can do about that apart from adding the Peacock feather to the tiles, which I think cheapens it a little.

Meaning no disrespect to the creators, but I think they should be cheapened a little. As they are they are indistinguishable from official escalations, which they are not. I think people should be able to tell if they're playing an official mission or a custom mission (or an unofficially restored mission). But that's outside of this scope, we can continue on discord if you want.

RDIL:
Closing in favor of Yanri's work, xref https://discord.com/channels/826809653181808651/1089161116527108157/1445777526680059924

PR #643, opened 2025-12-03
Update locale.json // german

sleeper286:
Added german translation for the latest Escalations and added missing objective translations from older Escalations


RDIL (approved):
Thank you!

PR #644, opened 2025-12-03
Add peacockContent option to control presence of peacock-exclusive content.

Yanrishatum:
## Scope

* Adds `peacockContent` option flag with supported values of `ENABLED` (default), `NOXP` and `DISABLED`
* In `NOXP` - challenges tagged with `peacock` will award no mastery XP to preserve (lol) mastery progression balance parity of official servers.
* When `DISABLED` - removes the challenges and escalations exclusive to Peacock outright.
* New challenges that are exclusive to Peacock now **should** have a `peacock` tag. Including challenge packs.
* New `type` field for `MissionManifest.Peacock` field: Currently only supports value of `peacock` to indicate that manifest is peacock-exclusive content.
  * New escalations/missions exclusive to peacock now **should** have that field set to `peacock`.
* Updated escalation guidelines to include tagging requirement.
* Misc: If challenge pack has no escalations it is now hidden in the `User Profile -> Challenges -> Challenge Packs` menu.

I've gone through stuff I know of, but I may have missed some challenges or escalations that are peacock-exclusive but not marked.

`type` field later can be utilized to filter other types of content that is missing on official servers, like content restoration. 

Supersedes by #641 as a more generic user-controlled option via `peacockContent=NOXP`.

## Test Plan

1. Check if all peacock exclusive challenges have proper `peacock` tag added.
2. Check if all escalations have `Peacok.type` set to `"peacock"`.
3. Set `peacockContent` to `NOXP`
4. Ensure all peacock-exclusive challenges have no XP awards.
    * If it has one - it was not tagged with `peacock` tag.
5. Set to `DISABLED`
6. Ensure all peacock-exclusive content is gone: challenges, escalations (inc. one in ET arcade).

Optional:

1. Set to `ENABLED`
2. Complete a few challenges and do progress/complete some escalations.
3. Set to `DISABLED`
4. Log in into game
5. Set to `ENABLED`
6. Ensure that progression was not lost.

## Checklist

### DIFF
--- a/components/candle/challengeService.ts
+++ b/components/candle/challengeService.ts
@@ -27,7 +27,7 @@ import type {
     GameVersion,
     MissionManifest,
     PeacockLocationsData,
-    PeacokcContentMode,
+    PeacockContentMode,
     RegistryChallenge,
     Unlockable,
     UserProfile,
@@ -321,7 +321,7 @@ export abstract class ChallengeRegistry {
 
         const set = locationMap.get(groupId)!
 
-        const peacockContent = <PeacokcContentMode>getFlag("peacockContent")
+        const peacockContent = <PeacockContentMode>getFlag("peacockContent")
 
         for (const challenge of challenges) {
             challenge.inGroup = groupId
--- a/components/controller.ts
+++ b/components/controller.ts
@@ -35,7 +35,7 @@ import type {
     MILLocations,
     MissionManifest,
     PeacockLocationsData,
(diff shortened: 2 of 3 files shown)


suanjiansalt:
> With how inventory currently works - when content is disabled - those items will be treated as "not in any reward pool" and thus available by default.

In any case it's not official-like. Unless you just mechanically define what the concept of "Peacock exclusive" is.

PR #652, opened 2025-12-14
fix(inventory): TOKEN_OUTFIT_WET_SUIT not showing in stashpoints

### LINKED ISSUE #650: TOKEN_OUTFIT_WET_SUIT no longer shows when selecting a starting suit
In the Hawkes Bay mission, if you select a different suit instead of the Tactical Wetsuit in the loadout screen, no longer use the Tactical Wetsuit afterward.

AnthonyFuller:
Fixes #650 
* Respect the "actual" LOCATION_NEWZEALAND default suit unlockable instead of using TOKEN_OUTFIT_WET_SUIT.
* No longer return stashpoint template for H3, add H2's template (modified to exclude TOKEN_OUTFIT_WET_SUIT in LOCATION_NEWZEALAND).

### DIFF
--- a/components/loadouts.ts
+++ b/components/loadouts.ts
@@ -210,7 +210,7 @@ export class Loadouts {
             loadout = dl.data[sublocation?.Id] ?? {}
         }
 
-        // SILLY HACK FOR OLD SAVES: For LOCATION_NEWZEALAND, if the saved suit is TOKEN_OUTFIT_WET_SUIT, replace
+        // SILLY HACK FOR OLD (< 8.6.0) SAVES: For LOCATION_NEWZEALAND, if the saved suit is TOKEN_OUTFIT_WET_SUIT, replace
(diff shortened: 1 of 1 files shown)


grappigegovert (approved):
Looks all good, only the comment in loadouts.ts took me some time to understand that it is about saved loadouts made in older versions. (which is what it is, right?)

components/loadouts.ts
```diff
@@ -210,6 +210,14 @@ export class Loadouts {
             loadout = dl.data[sublocation?.Id] ?? {}
         }
 
+        // STUPID HACK: For LOCATION_NEWZEALAND, if the selected suit is TOKEN_OUTFIT_WET_SUIT, replace it
```
grappigegovert: I would change 'selected' to say 'saved' instead.

RDIL (approved):

PR #658, opened 2025-12-26
refactor(profile): respect SubLocation.Properties.IsHidden when filtering sublocations

AnthonyFuller:
As title, we should've always done this, but we were probably rushing.

### DIFF
--- a/components/menus/destinations.ts
+++ b/components/menus/destinations.ts
@@ -290,7 +290,7 @@ export function createLocationsData(
         const sublocation = locData.children[sublocationId]
 
         if (
-            sublocationId === "LOCATION_TRAPPED_WOLVERINE" ||
+            !sublocation.Properties.CreateContractId ||
             sublocation.Properties.IsHidden
         ) {
             continue


components/menus/hub.ts
```diff
@@ -193,15 +193,16 @@ export function getHubData(gameVersion: GameVersion, userId: string) {
     }
 
     for (const child in locations.children) {
+        const location = locations.children[child]
+
         if (
             child === "LOCATION_ICA_FACILITY_ARRIVAL" ||
```
RDIL: is this still the case?

AnthonyFuller: Hiding arrival? Yes. Otherwise, I'm not sure what you're asking.

AnthonyFuller: Yes, IsHidden was added specifically for the Evergreen locations. Hence the comment about the HideProgression key (which is set to true for Arrival).

AnthonyFuller:
Well, issues with the classics was reported by Kotti with them not showing on Holiday Hoarders, while they are shown on official.

PR #661, opened 2025-12-28
fix: Loadouts being empty on brand-new profiles

RDIL:
## Scope

In 11b437320e4f0e7a266ecef634aa36df6ce62356 we did some refactoring, but because of how it ended up, this code path never gets hit when using loadout profiles.

Fixes an issue reported on Discord where new profiles would start with empty loadouts.

## Test Plan

After the fix, create a new profile and ensure the items are properly selected:

Screenshot 2025-12-28 134438

## Checklist

--------
#### General

--------
#### Testing


AnthonyFuller (approved):
Ah, just compared it with the previous logic. This was my bad when making the new function.

PR #678, opened 2026-03-21
refactor: Make inventory class-based

RDIL:
## Scope

Two big changes here:
1. inventory is a class now, so the stateful parts are not stored on a module level
2. controller instance can be overridden for tests, so that code which hasn't yet been migrated to the IoC model can use the intended controller instance.

## Test Plan

- Start the server, verify that logging in still works

## Checklist

--------
#### General

--------
#### Testing


alex73630 (approved):
Overall LGTM

Of course it would be better to have it checked against tests but I know that this change is required to get tests running.

PR #679, opened 2026-03-22
feat: allow for authenticating Steam users without official

AnthonyFuller:
As title. Validates using the app ticket.

Pretty much done aside from the backend implementation (and final polish).


AnthonyFuller:
Okay, so updating node fixed esbuild failing. Good to know.
EDIT: Apparently not it just successfully built? Very confusing - I must've opened the previous run or something...

AnthonyFuller:
This is technically ready to merge, but we should wait for @RDIL to update the privacy policy first. (I would add a request changes review to block it, but I cannot do that on my own PR)

Also if you couldn't tell, I can't run the tests locally because I get an error: `SyntaxError: Octal escape sequences are not allowed in strict mode.`

AnthonyFuller:
Backend is now up and running and is able to handle the auth requests successfully.

PR #693, opened 2026-05-19
Add support for 410 responses in OAuth requests

grappigegovert:
Adds support for handling 410 responses coming from IOI's servers when sending an oauth request, and also adds support for sending 410 responses of our own in oauth requests.

A response of `410 Gone` is sent by IOI's servers when sending a request to `/oauth/token` with `pId` set to a profile id that is not linked with the external user id used for the authentication.
When the game receives such a response, it will retry the oauth request with `pId` set to an empty string.

## Scope

I have added this retry logic (to mimic the game client) in `officialServerAuth.ts`, and changed some bits in `webFeatures.ts` to use the correct new profile id returned by IOI's servers.
If a user is logged into peacock with a profile id that does not match their profile id on official servers, this will allow them to still do a progression transfer instead of hitting an error. (This can happen for example when logging into peacock without having logged into official servers before)

Peacock can now also send a 410 response when `pId` in the oauth request does not match the profile id linked to the external user.
If for example a user now switches steam accounts and logs into peacock again, a new profile will be created instead of linking the second steam account to the same user profile used by the first steam account. Before this fix both steam users would be linked to the same peacock profile

## Test Plan

I have done some testing with this of course and I think it all works correctly but I didn't want to straight push this to master without someone else looking at it first.

I thought about adding an `officialUserId` field to user profiles so we only have to do the retry once, but because the large majority of users will never need this (their official user id is just the same as their peacock profile id) I'm not sure; I'd like to hear your thoughts

--------
#### General

--------
#### Testing


RDIL (approved):
LGTM! just needs to fix the failing tests

AnthonyFuller (approved):
Tests are fixed (the mock returned an empty string instead of the pId). I think this change LGTM but haven't done any extensive testing.

grappigegovert:
I'll go ahead and merge it then, thanks for fixing the tests

PR #712, opened 2026-06-28
Deprecate modIsInstalled in favour of new SMFv3-style shims

atampy25:
Introduces `modEnabledForUser` and `modEnabledForGame` from #710 and deprecates the existing `modIsInstalled` in favour of the new functions. SMFv2 does not actually provide the necessary data to check the mod version, or information for a specific game, so these functions currently have the exact same behaviour as `modIsInstalled`. Nonetheless, they have the same signature as the ones in #710, so will help get developers in the habit of specifying the correct information.

--------
#### General

--------
#### Testing

### DIFF
--- a/components/smfSupport.ts
+++ b/components/smfSupport.ts
@@ -30,6 +30,7 @@ import path, { basename, join } from "path"
 import { readFile } from "fs/promises"
 import { menuSystemDatabase } from "./menus/menuSystem"
 import { parse } from "json5"
+import { deprecated } from "./utils"
 
 type LastServerSideData = SMFLastDeploy["lastServerSideStates"]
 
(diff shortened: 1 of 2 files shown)


components/smfSupport.ts
```diff
@@ -216,11 +221,54 @@ export class SMFSupport {
      *
      * @param modId The mod's ID.
      * @returns If the mod is available (or the `overrideFrameworkChecks` flag is set). You should probably abort initialisation if false is returned.
+     * @deprecated since v8.9.0, use `controller.smf.modEnabledForUser` or `controller.smf.modEnabledForGame`
      */
     public modIsInstalled(modId: string): boolean {
```
RDIL, line 227: we added a deprecation warning helper in a recent PR, maybe we should add that here as well so you can see in the console there is a problem?

RDIL (approved):
LGTM modulo add deprecation warning via the new helper in utils

components/smfSupport.ts
AnthonyFuller: Agreed!

atampy25:
Should be ready to merge

Issue #596, opened 2025-06-17
Game Crash when loading into missions


charc0al:
Platform: Steam and Epic
OS: Windows 11

Getting game crashes intermittently (about ~10% of the time) when loading or unloading a level. Analyzed MANY crash dumps and almost all of them are failing here

ExceptionAddress: 000000014138a96c (HITMAN3!cat::ClientEasyHandshake::ProcessAnswerWithIdentity+0x000000000031289c)

I get this same crash on both the Steam(no mods) and Epic(SMF+ZHM SDK) versions of WoA, on the current version 3.230. Installed the game completely fresh on Steam, with no mods, and a fresh install of peacock 8.1.0 and nothing else, and still get these intermittent crashes. When I don't use peacock I don't get any game crashes. There doesn't seem to be any pattern as to which maps will crash when loading. Looks like something in the client handshake response is causing an exception.

![Image](https://github.com/user-attachments/assets/35e28624-b367-410c-8a9c-a8678093e2fb)

RDIL:
You're getting this on an unmodified version of the game, with Peacock?

I'd be very surprised if this is a Peacock-caused crash. Looks a lot more like Windows or the game itself.

charc0al:
It's weird that if it's an SSL error it wouldn't crash every time. I thought it was crashing because of mods / SDK so I tried a completely clean install from Steam, and the latest release of peacock 8.1.0 with no plugins, still get crashes on ~10% of level loads. I'm thinking it might be a timing issue where the game isn't expecting to get a response back so quick (the server wasn't intended to be local) and maybe hitting some code before it's initialized. I'll try inserting some delays in a few functions and see if that affects anything

VoodooHillbilly:
I've been getting random crashes during loading for several months. It doesn't occur when I have a vanilla Steam install on IOI servers without the SDK. The crashes start again when I use the SDK to use the Peacock server. The crash_metrics.dat and piepieonline's CrashDebugHelper.dll didn't show a reason for the game crashing. So I originally thought it was due to something in the SDK itself.

I've seen several other mentions of random crashes during loading on the Glacier Modding Discord, usually in the #sdk channel. There was another report of it yesterday in #peacock-chat when I linked this Issue.

charc0al:
[...]

PR #721, opened 2026-08-17
Fix crash that can happen while loading into missions from the menu

hdr88:
## Scope

This change will fix crash that can happen while loading into missions from the menu. Relevant issue:
https://github.com/thepeacockproject/Peacock/issues/596

## Test Plan
After countless hours of trial and error, i narrowed the source of the crash to the lines 4-6 in static/peacock-menu/index.json:
        "datavalues": {
            "@global.CacheBuster": "0"
        },
Removing this, the game no longer seems to crash during loading. Verified by loading into missions over and over (100+ at least). Without this fix, the game crashes something like 1 out of 15 times or so when loading a mission. Maybe i just got unbeliavably lucky though, so i would appreciate others trying this fix.
I also verified that the peacock menu in settings still work, so not sure what this CacheBuster is used for, other than making the game crash.

### DIFF
--- a/static/peacock-menu/index.json
+++ b/static/peacock-menu/index.json
@@ -2,11 +2,11 @@
     "$datacontext": {
         "in": "$.",
         "datavalues": {
-            "@global.CacheBuster": "0"
+            "CacheBuster": "$.@global.CacheBuster"
         },
         "do": {
             "$include": {
-                "$path": "$formatstring menusystem/pages/peacock-menu/options.json@{$.@global.CacheBuster}"
+                "$path": "$formatstring menusystem/pages/peacock-menu/options.json@{$.CacheBuster}"
             }
         }
     }


AnthonyFuller (dismissed):
This is not exactly what I'd describe as a desirable fix. The "CacheBuster" is exactly that, the game caches JSON files retrieved from the server and the CacheBuster forces it to be re-requested. I don't think this is just a Dev feature, I believe it's partly needed for the general operation of the in-game menu.


I'm not dismissing this finding outright since it makes sense that this is causing a crash when it's destroyed and re-created, but I have definitely experienced a somewhat similar crash on official.

LukaMegurine6:
> ## Scope
> 
> This change will fix crash that can happen while loading into missions from the menu. Relevant issue: #596
> ## Test Plan
> 
> After countless hours of trial and error, i narrowed the source of the crash to the lines 4-6 in static/peacock-menu/index.json: "datavalues": { "@global.CacheBuster": "0" }, Removing this, the game no longer seems to crash during loading. Verified by loading into missions over and over (100+ at least). Without this fix, the game crashes something like 1 out of 15 times or so when loading a mission. Maybe i just got unbeliavably lucky though, so i would appreciate others trying this fix. I also verified that the peacock menu in settings still work, so not sure what this CacheBuster is used for, other than making the game crash.
> ## Checklist

I'm going to test that, i'm not 100 % sure too but if it's reduce, it's could be already great tbh

LukaMegurine6:
The fix works...
I'm happy but very suprised too
Before I couldn't load the bank, Haven and of course FV without crashing at least once, now I can load everything in one session...wow
I restarted the game 10 times, I was able to launch FV each time and the other destinations without any crash
And FV mission to safehouse don't crash too

It's does not result memory crash (ofc) but it's seem to work just fine !

LukaMegurine6:
EDIT : It's not result all crashes, probably someone else need to test the fix.

hdr88:
> This is not exactly what I'd describe as a desirable fix. The "CacheBuster" is exactly that, the game caches JSON files retrieved from the server and the CacheBuster forces it to be re-requested. I don't think this is just a Dev feature, I believe it's partly needed for the general operation of the in-game menu.
> 
> I'm not dismissing this finding outright since it makes sense that this is causing a crash when it's destroyed and re-created, but I have definitely experienced a somewhat similar crash on official.

Even without the changes i did, if i change some peacock setting in the menu, i do not see the game sending a request for a new JSON file at all. What it does, it sends a SubmitSemEmail request which is used to getAllFlags. Also i can see that in the response of the SubmitSemEmail request, there is a CacheBuster field and its value is a date, but it is not clear to me whether the game uses this for anything.
I validated, even with my changes, the peacock menu behavior is exactly the same like i described above.

LennardF1989:
@hdr88 Could you test this commit? https://github.com/thepeacockproject/Peacock/commit/57e44fd43d0ab583f84d8684ff621cde14d0ad2b

Branch [lennard/cachebuster-fix](https://github.com/thepeacockproject/Peacock/compare/lennard/cachebuster-fix) if you just want to clone and go.

Do **not** apply it on top of the changes in the PR.

hdr88:
> @hdr88 Could you test this commit? [57e44fd](https://github.com/thepeacockproject/Peacock/commit/57e44fd43d0ab583f84d8684ff621cde14d0ad2b)
> 
> Branch [lennard/cachebuster-fix](https://github.com/thepeacockproject/Peacock/compare/lennard/cachebuster-fix) if you just want to clone and go.
> 
> Do **not** apply it on top of the changes in the PR.

I tested the commit and it works also, loaded into missions many times, game does not crash.
The request URLs for peacock menus do change with this commit compared to master, they no longer have 0 in them. For example, they look like this:
/peacock-menu/options.json@?sv=2018-03-28&ver=h3
instead of this
/peacock-menu/options.json@**0**?sv=2018-03-28&ver=h3
On master, that 0 comes from assigning value to @global.CacheBuster in static/peacock-menu/index.json, which is not safe to do it seems.
Without explicitly assinging 0 to @global.CacheBuster, it seems like it does have any value, at least it did not have during my testing.

LennardF1989:
Thank you for confirming, the missing "0" after the @ was intentional. There are ways to get it back, but it doesn't add anything. 

My branch would therefor be the preferred fix to the crash issue, since it does not remove the cache buster functionality during development.

hdr88:
I retested the game after the new update (3.280), but the behavior is unchanged.
The game can still crash during loading without the change in the PR.
Did not observe any crash during loading after the change.

RDIL (approved):

PR #724, opened 2026-09-03
Fix FC ID added to the wrong FC challenge track

hdr88:
## Scope

While playing the newly added FCs, i have run into an issue that 1 of them has been counted on the wrong FC challenge track.
To reproduce the issue:
Start any FC, note its associated FC challenge track (H1+H2 or H3), but do not finish it, quit to menu.
Now start a different FC belonging to a different FC challange track and finish it.
The completion of the FC will also count for the non associated challenge track, when it should not.

Root cause:
While a FC challenge track is in state State_ValidContract, it will only check if the running contract is featured, but does not check whether the location of the FC is correct or not.

I have changed the FC challange tracks, so this issue does not happen anymore.

## Test Plan

Verified by playing some FCs. Challenge tracks still work, but can no longer reproduce the above issue.


hdr88:
After some extra testing i realized my first solution was not backward compatible, because if a challenge was already in the State_ValidContract state, it did not progress with it.
So instead i just updated the State_ValidContract state to check for the location as well, not just the Start state.

grappigegovert:
Alright, I just did some testing and I thought that peacock was doing something wrong with the challenge scope here, because the challenge definition is straight from IOI, so the problem should be somewhere else.
However, it turns out that this exact same bug is present on IOI's servers.
I would think that because the challenge scope is `Hit` the progression is stored per-contract, but not only does Peacock just store the progression profile-wide, IOI's servers do the same thing 🤷

hdr88:
I see, i suppose it would be a bad idea to differ from IOI, even if it is a bug. I am good to close the PR then, but let me know what you suggest.

RDIL:
Yeah, good fix, but we shouldn't diverge like this.

PR #725, opened 2026-09-10
feat: add "Challenge Packs" category to the Career -> Challenges page

AnthonyFuller:
As title. Also adds an option (`hideGhostModeChallengeTile`) that allows one to hide the tile in H3 only.

Smidge bit hacky, but it works. More template json than I would've liked though.
Et voila
Translations of "Challenge Packs" were taken from HITMAN 2.


grappigegovert:
Looks good to me 👍 
I don't like either how much templates have to be edited but I think you kinda have to. I was thinking you could maybe check if the peacock strings mod is installed and then do the changes in the mod, so other mods can also edit the templates, but that might make the whole thing even more complicated for no reason.

Making `IsPack` false to hide the versus pack works neatly, but it looks very hacky, I think it'll be confusing for people reading the code in the future.
Maybe you can change that to a conditional operator on getFlag, something like
```ts
getFlag("hideGhostModeChallengeTile") ?
    !(gameVersion === "h3" && id === "versus")
    : true
```
I think that way it's more clear why you're manipulating IsPack at all

AnthonyFuller:
Totally agree, done. I think I was too tired to see that lmao

grappigegovert (approved):

PR #727, opened 2026-09-12
fix: Update scanner for 3.270.1

RDIL:
## Scope

The compiler output changed drastically enough that it no longer was working.

## Test Plan

It works now:

Screenshot 2026-09-12 at 08 53 51


grappigegovert (approved):
Looks alright to me. I'm not super familiar with arm assembly, so I'm just assuming your math for the branch offset is correct, but the changes and explanations in comments look good.

Just to check, is new pattern for 3.270.1 and 3.280.0 (the latest version)?
The patcher already has hardcoded offsets for 3.270.1 but I assume that was just to bypass the non-working scanner for that version.

RDIL:
Yeah it technically changed in the last version, I'll update the comment

RDIL:
Oops the comment already is correct. Yeah it changed in the last version and stayed in the same in the current version.

PR #730, opened 2026-09-13
update autosplitter reset rule to a flat +1s

Rekt05:
## Scope

Updated the autosplitter reset logic to be a 1s minimum for all resets, including those in the cutscene, and removed the +10s logic as it is no longer a part of the [speedrun.com rules](https://www.speedrun.com/hitman_woa?rules=category#:~:text=in%20correct%20order.-,Timing,-While%20full%2Dgame)
[src post about no 10s penalty](https://www.speedrun.com/hitman_woa/news/jxnqklxk)
[src post about 1s minimum penalty](https://www.speedrun.com/hitman_woa/news/z3zmmq6d)

## Test Plan

I did a run with livesplit and ensured that all resets have a minimum +1 for resets below 1s and resets in the cutscene, as well as ensuring bangkok doesnt add +10 for resets anymore


fu5ha (approved):
Cool, looks good to me

